PR #35, opened 2015-08-04
Create zh_CN.lang

Joccob:


maxanier:
Thanks for translating, but we moved the translations to https://crowdin.com/project/vampirism/zh-CN. I think you can simply upload the translated file there, if not I will do it.

PR #89, opened 2016-07-03
Update to 1.10.2

insuusvenerati:
Uncomment apply from: 'gradle/deploy.gradle' from the build.gradle to fix Curse

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ server.properties
 usercache.json
 whitelist.json
 hs*
+/.idea/
+/classes/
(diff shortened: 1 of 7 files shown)


maxanier:
Thanks for updating, I unfortunately already had done it locally, but forgot to commit and push it, because I had too much else to do. 

I will release a version on curseforge later today

insuusvenerati:
No worries. Only took a minute to do and I wanted to do it locally cause I needed to use it for 1.10.2

PR #188, opened 2017-06-09
Adds an option to disable the vampire hud

tyra314:
This adds an config file entry to disable the vampire hud. The hud
causes severe graphics errors, when shaders are used. While this doesn't
fix it, it provides a workaround.


maxanier:
Will modify the config comment through to make clear that is deactivates the red and yellow overlay, not the other GUI elements

tyra314:
Thanks for adapting and merging 👍

PR #254, opened 2017-12-22
Update zh_CN.lang

ETStareak:
Hello!
I'm a translater in Chinese,and this is the newest zh_CN.lang for your mod,you can use it in many versions.


maxanier:
Hello.
Thank you very much for translating.
However, translations are handled via Crowding to keep them separate from the code.
I have uploaded you translation there. The main language file is no almost complete.

The new translations will be included in the next versions.

By the way. You can find the translations here https://crowdin.com/project/vampirism/zh-CN

maxanier:
Let me know if there are any issue with the translation as soon as the next version has been released.

Closing this since the translation have been uploaded to crowdin

PR #321, opened 2018-07-01
Biome Weight from config is ignored.

erebus42:
Changed it to use the value from BalanceGeneral.VAMPIRE_FOREST_WEIGHT

This is my minecraft related PR, but this seemed like an easy fix.

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/biome/BiomeGenVampireForest.java
+++ b/src/main/java/de/teamlapen/vampirism/biome/BiomeGenVampireForest.java
@@ -3,7 +3,6 @@ package de.teamlapen.vampirism.biome;
 import de.teamlapen.vampirism.VampirismMod;
 import de.teamlapen.vampirism.blocks.VampirismFlower;
 import de.teamlapen.vampirism.core.ModBlocks;
-import de.teamlapen.vampirism.config.BalanceGeneral;
 import de.teamlapen.vampirism.entity.EntityBlindingBat;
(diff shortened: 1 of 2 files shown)


maxanier:
Any reason you closed this? Looks like a valid bug and a working fix.

erebus42:
I wasn't sure, and the failed test kinda scared me off. I can re-open it if
you'd like.

erebus42:
Re-opened.  Also was this the correct branch? It looks like the build failure was unrelated to my change.

erebus42:
I actually left the name change out, since it was referenced elsewhere. the comment is incorrect.  This is just the use config weight change.

maxanier:
The branch is correct.
However, the failed build is related to this change.
The balance config system creates the balance class objects only at startup and loads the values from the config files during preinit. Therefore all config values are non final and non static (I know the all caps are misleading), so you cannot access the biome weight by `BalanceGeneral.VAMPIRE_FOREST_WEIGHT` but you need to do `Balance.general.VAMPIRE_FOREST_WEIGHT` (all initiated balance objects are stored in the Balance class in static variables).
Additionally, the value might not be available when the BiomeGenVampireForest class is loaded so it should not be assigned to a final static variable (even though it is likely that the class is only loaded after preinit).

I would suggest completely removing the GEN_WEIGHT field and just using `Balance.general.VAMPIRE_FOREST_WEIGHT` in the ModBiomes initialisation.

[I will do this if you want]

maxanier:
Thanks for reporting and fixing

PR #332, opened 2018-08-02
1.12 

Cheaterpaul:
-added colored vampirecloak
-blood grinder takes values from txt file (vampirism:blood_values_grinder)
      -support for mods
-if bat_duration is unter 10% show visual notification


src/main/java/de/teamlapen/vampirism/client/gui/VampirismHUDOverlay.java
```diff
@@ -44,303 +46,332 @@
...
+			renderFullTick--;
+		}
+
+	}
+
+	private void setColorForBatDuration(IFactionPlayer player) {
+		float batPercentage = ((IVampirePlayer) player).getActionHandler().getPercentageForAction(VampireActions.bat);
+		if (batPercentage < 0.1F && batPercentage > 0.0F) {
+			screenBottomColor = 0xcc7067f9; // change color
+			screenBottomPercentage = (int) (0.1F - batPercentage * 10);
```
maxanier: I assume screenBottomPercentage should go from 0 to 100. Since batPercentage can have a range between 0.1 and 0). So I would propose changing this to `(int) (100.0F*10*(0.1F - batPercentage))`

maxanier:
Superseded by #334

PR #341, opened 2018-08-14
Vampircloak

Cheaterpaul:
more fitting vampire cloak model & texturemaps


maxanier:
Looks good.
But can you rename the EnumColor to something more cloak specific like "EnumCloakColor" and move it into a non API package, maybe even into the ItemCloak class. I don't think a third party mod needs access to it. Or is there a specific reason you put it there?

Cheaterpaul:
updated

maxanier:
Will merge once I am home. Thank you again :)

Might push the first 1.5 alpha soon

maxanier:
BTW: Are the cloaks already craftable? If not did you already think about what would be the best way to get them.
Because they are quite cool, I think it would be reasonable to treat them like an expensive item. So either the recipe requires some rare items, or we add them as rare loot in chests or from mobs.

Cheaterpaul:
Maybe made with wool, diamonds and a vampire soul with a kind of energy object (like a living cloak -> Poisones to Hunter?).
The energy item could be collected from loot if it is a new item of vampirism.

maxanier:
If you prefer a recipe that sounds ok. However, I think we don't really need a energy object (unless we want to implement more of that). Maybe we could use a human heart or even a pure blood bottle for that (in addition to some wool, a diamond and some dye maybe)

Cheaterpaul:
Yes, that's what I thought about

maxanier:
Nice.
One last thing :)

Maybe you can add a hint "poisonous to hunters" like here: https://github.com/TeamLapen/Vampirism/blob/53475bbacccef2e0197895bede7892d821627119/src/main/java/de/teamlapen/vampirism/items/VampirismHunterArmor.java#L124

(no need for that 1.11 helper stuff, just put it directly in the addInformation method https://github.com/TeamLapen/Vampirism/blob/53475bbacccef2e0197895bede7892d821627119/src/main/java/de/teamlapen/vampirism/items/VampirismHunterArmor.java#L55)

maxanier:
I modified the recipes slightly because the vampire soul is exclusively available to hunters

PR #343, opened 2018-08-15
Hunter tent spawn limit

Cheaterpaul:
added spawn limit per day.
now: spawn limit is 5
edit at in [BalanceGeneral](https://github.com/Cheaterpaul/Vampirism/commit/bb05a19003ddc941f8e720f9c5b828e647eb36ad)


src/lib/java/de/teamlapen/lib/lib/util/SimpleSpawnerLogic.java
```diff
@@ -30,6 +32,9 @@
     private int maxNearbyEntities = 4;
     private int spawnRange = 4;
     private int spawnDelay = 20;
+    private int spawnedToday = 0;
+    private float spawnedLast = 0;
```
maxanier: Why is this a float? Looks like we are storing integer (or longs) in here

src/lib/java/de/teamlapen/lib/lib/util/SimpleSpawnerLogic.java
```diff
@@ -153,11 +168,13 @@ public void updateSpawner() {
...
                         onSpawned(entity);
-                        flag = true;
+                        flag1 = true;
                     }
                 }
-                if (flag) {
+                if (flag1) {
                     this.resetTimer();
+                    this.spawnedToday++;
+                    this.spawnedLast = getSpawnerWorld().getWorldTime();
```
maxanier: Not sure how worldTime behaves (it might become larger than 24000). Furthermore worlTime does not change if the day night cycle is disabled.
So just to be sure I would use getTotalWorldTime % 24000 here (as well as above).
Should probably be worldTime % 24000. Otherwise the

src/lib/java/de/teamlapen/lib/lib/util/SimpleSpawnerLogic.java
```diff
@@ -126,7 +134,14 @@ public void updateSpawner() {
                     return;
                 }
 
-                boolean flag = false;
+                if ((getSpawnerWorld().getWorldTime()) % 24000 < this.spawnedLast) {
```
maxanier: see below

maxanier (changes requested):
Looking good, but I have commented on a few things that probably have to be changed

maxanier (approved):

PR #344, opened 2018-08-15
new semi Invulnerable action

Cheaterpaul:
added action
-balance in [BalanceVampireAction](https://github.com/TeamLapen/Vampirism/blob/1.12/src/main/java/de/teamlapen/vampirism/config/BalanceVampireActions.java)


src/main/java/de/teamlapen/vampirism/config/BalanceVampireActions.java
```diff
@@ -99,6 +99,15 @@
...
     public double DARK_BLOOD_PROJECTILE_INDIRECT_DAMAGE;
 
+    @DefaultBoolean(value = true, name = "invulnerable_enabled")
+    public boolean HALFINVULNERABLE_ENABLE;
+    @DefaultInt(value = 20, minValue = 0, name = "invulnerable_cooldown")
+    public int HALFINVULNERABLE_COOLDOWN;
+    @DefaultInt(value = 30, minValue = 1, name = "invulnerable_duration")
+    public int HALFINVULNERABLE_DURATION;
+    @DefaultDouble(value = 0.5D, minValue = 0.0D, comment = "border for damage stop")
+    public double HALINVULNERABLE_BORDER;
```
maxanier: Please rename to *_THRESHOLD and adjust the comment accordingly. Maybe make it more specific (mention that this is relative to the max health)

src/main/java/de/teamlapen/vampirism/player/ModPlayerEventHandler.java
```diff
@@ -124,6 +126,13 @@ public void onLivingAttack(LivingAttackEvent event) {
             if (!FactionPlayerHandler.get((EntityPlayer) event.getEntity()).onEntityAttacked(event.getSource(), event.getAmount())) {
                 event.setCanceled(true);
             }
+
```
maxanier: This could be moved to VampirePlayer#onEntityAttacked. But I guess it is fine here too

src/main/java/de/teamlapen/vampirism/player/vampire/actions/HalfInvulnerableAction.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    @Override
+    public boolean isEnabled() {
+        return Balance.vpa.HALFINVULNERABLE_ENABLE;
+    }
+
+    @Override
+    protected boolean activate(IVampirePlayer playerIn) {
+        VampirePlayer player = VampirePlayer.get(playerIn.getRepresentingPlayer());
```
maxanier: You can just cast IVampirePlayer to VampirePlayer.
The argument is IVampirePlayer so mods using the API can make use of it (since they cannot access VampirePlayer), but the actual object passed is always an instance of VampirePlayer

src/main/java/de/teamlapen/vampirism/player/vampire/actions/HalfInvulnerableAction.java
```diff
@@ -0,0 +1,83 @@
...
+ */
+public class HalfInvulnerableAction extends DefaultVampireAction implements ILastingAction<IVampirePlayer>{
+
+    public HalfInvulnerableAction() {
+        super(null);
+    }
+
+    @Override
+    public int getCooldown() {
+        return Balance.vpa.HALFINVULNERABLE_COOLDOWN;
```
maxanier: Probably want to do this in seconds, so times 20

maxanier (changes requested):
Maybe lower the default value to 20, since the skill might be incredibly strong in modded Minecraft.
Therefore it might also be reasonable to add maybe a slowness boost. Maybe it would be even better to make the effect use blood instead. So for every consumed hit you loose maybe 4 blood.

src/main/java/de/teamlapen/vampirism/player/vampire/BloodStats.java
```diff
@@ -41,12 +42,17 @@ public int addBlood(int amount, float saturationModifier) {
 
     @Override
     public boolean consumeBlood(int a) {
```
Cheaterpaul, line 44: return true everytime?

maxanier: :see_no_evil: good point

maxanier: but no need, to create a new method, just fix the old one :D

maxanier:
It looks like the changes to VampirePlayers have not been commited/pushed. At least I cannot find the part that actually reduces the damage

Cheaterpaul:
seems like i told git to ignore the class

maxanier (approved):

maxanier:
Looks good now.
I did remove a log and adjusted the balancing values as we discussed.
Furthermore I made sure the slowness effect is removed if the action is canceled prematurely.

Also, I renamed the skill to "Damage Limiter", since half invulnerability might be a little bit confusing. I hope you do not mind. We can change it back if you want

Issue #312, opened 2018-06-23
Make sucking blood take time


maxanier:
Idea by CptAlfr

Instead of instantly sucking out the blood of an entity, make the process take time.

While sucking blood the victim would be frozen.
The sucking process could be interrupted by dealing damage to the vampire.

By varying the  duration of the sucking process the vampire could decide how much of the victims blood they would like to take.

Also, cool particles 

Main downside: More prone to bugs as this would be a process over several ticks

ThugPug43:
I think a screen effect would also be fitting. When your blood is being sucked, your screen has a slowly increasing black, or red, vignette. (maybe fade in quickly as black then fade to red the closer you are to having your blood sucked.

PolarisHD:
wouldn't it be easier to give slowness to both the vampire and the vampire's victim instead of making the action itself take a long time?

maxanier:
Would be a lot easier but rather pointless I thin. It would not add much immersion but mostly annoy people

Sh0wYaH0wda:
what about just make it the longer you feed the more blood you get

maxanier:
Definitively (Of course limited by to total amount the victim has)

maxanier:
Just noticed there was a similar issue #133 , moving the comment here:

> Some vampire mechanics feel a bit off. Mainly the feeding. People have commented on the current feeding system and would like to see some changes, the first being made to the way that a vampire feeds. Rather than being an instant thing, it should be a slow drain over time. Let's say 5 seconds for a full drain that would kill a villager (and/or possibly turn them). Another suggested idea would be a "blood meter" on the side of the screen that would fill if your health and hunger were both full after feeding. This meter will be drained to heal the player and restore their hunger when lost and should last about 20 minutes from full at all levels before it is completely empty. Higher level vampires should still drain hunger faster, but have a larger blood meter and get more blood from feeding to compensate so that at any level one villager death will suffice (as long as they dont need to fight). Vampire abilities should also drain the blood bar to be used. But this should be a fixed value that only changes if the ability is improved somehow. This means that high-level vampires can use their abilites more in between feeding, but still have to suffer the cooldown whereas low-level vampires can only use it once or twice. This would also be a bit more of an encouragement to take blood bottles with 
you...

This additionally suggests using the blood in more ways (e.g. for abilities) which is a good idea, however, I would skip the blood meter for now

tsheinen:
[...]

PR #355, opened 2018-08-27
[WIP] Feed over time

tsheinen:
As discussed in issue #312 i've been working on a continuous feeding effect.  This PR is a work in progress and I just made it so we can discuss it.


maxanier:
Functionality look good now, I would move the actual blood transfer to the end of the cycle though (so you have to feed for 20 ticks before you get the first chunk of blood).

The issue with the current code is that it does not work in multiplayer.

The server cannot check if a player presses a button (ModKeys#isDown) or the mouseover entity (Minecraft#objectMouseOver). Actually, it does not even know about these classes because they are not included in the MC server build, so the server crashes when trying to use them.
More about that on the Forge wiki: https://mcforge.readthedocs.io/en/latest/concepts/sides/

To avoid this we need the client to send an packet if we start feeding blood as well as if we stop (release key or move mouse away). The server needs to process the events and tell the client what is going on so it can render things accordingly.

The release key check could be done in ModKeys, the rest probably in VampirePlayer
```
if(isRemote()){
  //Client side
  
}
else{
  //Server side
}
```
This would handle the logical difference. We still need to make sure we are not using any Client only methods, so if we e.g. want to access the `objectMouseOver` in VampirePlayer. We need to add a method like `@Nullable Entity getMouseOverEntity` to `IProxy`. Make it return null in server proxy and the mouse over entity in client proxy

The server could update the client using the implemented update/sync setup:

Create a NBTTagCompound with the necessary information and call `sync(nbt,false)`
Implement the (client side) processing in `loadUpdate`.
Probably also want to add it  to `writeFullUpdateToNBT`.

(You could have a look on how we do the Fang type update. `setFangType` `loadUpdate` `writeFullUpdate`)

maxanier:
Forgot to mention how to access the proxy instance, just in case you did not find out by now:
`VampirismMod.proxy` gets you the `IProxy` instance (server side it will be `ServerProxy`, client side (and integrated server) it will be `ClientProxy`)

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -677,6 +719,11 @@ public void onUpdate() {
                     player.addPotionEffect(new PotionEffect(ModPotions.thirst, hunterEffect.getDuration(), hunterEffect.getAmplifier()));
                     player.removePotionEffect(MobEffects.HUNGER);
                 }
+                if (player.ticksExisted % 10 == 0) {
```
maxanier: We don't want to sync the victim id twice per second.
It should only be sent if it changes.

For ending the bite process, you could use the exiting `syncPacket` here, by writing the id to the exiting object and setting `sync` to true.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -691,10 +738,11 @@ public void onUpdate() {
                     sync(syncPacket, syncToAll);
                 }
             } else {
-
                 ticksInSun = 0;
             }
-
+            if (victim != -1 && player.ticksExisted % 20 == 0) {
```
maxanier: We still need the bite timer.  `ticksExisted` does not start at 0 but at a "random" number if you start feeding. Which results in an inconsistent behavior how long it takes for the first blood transfer.

Probably want to move this block into the `level>0` part above too.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -707,6 +755,14 @@ public void onUpdate() {
             } else {
                 ticksInSun = 0;
             }
+            if (player.ticksExisted % 5 == 0 && victim != -1) {
+                spawnBiteParticle(victim);
```
maxanier: This is quite annoying.
The server already triggers a sound when the blood is actually transfered, that should be enough I think.
Alternative would be to do use actual blood particles flying from the biten entity to the player simlar to what the altar of  inspiration does https://github.com/TeamLapen/Vampirism/blob/218a791e30962cd2c5e650b14fc34ceca954b1c7/src/main/java/de/teamlapen/vampirism/tileentity/TileAltarInspiration.java#L82-L82

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1018,6 +1077,7 @@ protected void writeFullUpdate(NBTTagCompound nbt) {
         nbt.setInteger(KEY_EYE, getEyeType());
         nbt.setInteger(KEY_FANGS, getFangType());
         nbt.setBoolean(KEY_GLOWING_EYES, getGlowingEyes());
+//        nbt.setInteger(KEY_VICTIM_ID, victim);
```
maxanier: Just to be sure I would leave this active

src/main/java/de/teamlapen/vampirism/proxy/IProxy.java
```diff
@@ -24,4 +27,7 @@ default float getRenderPartialTick() {
     boolean isPlayerThePlayer(EntityPlayer player);
 
     void renderScreenFullColor(int ticksOn, int ticksOff, int color);
+
+    @Nullable
+    Entity getMouseOverEntity(EntityPlayer player);
```
maxanier: Any reason for the player argument? It only works on client side anyway and the argument is not even checked

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -216,15 +226,47 @@ public void biteEntity(int entityId) {
...
         }
         if (e != null && e instanceof EntityLivingBase) {
-
             if (e.getDistance(player) <= player.getEntityAttribute(EntityPlayer.REACH_DISTANCE).getAttributeValue() + 1) {
-                biteEntity((EntityLivingBase) e);
+                if (determineBiteType(((EntityLivingBase) e)) == BITE_TYPE.ATTACK) {
+                    biteEntity(((EntityLivingBase) e));
+                } else {
+                    victim = e.getEntityId();
+                    PotionEffect effect = new PotionEffect(Potion.getPotionFromResourceLocation("slowness"), 20, 7);
```
maxanier: We could use a custom potion effect here too, this could be used if a player is bitten to render their screen red

src/main/java/de/teamlapen/vampirism/client/core/ModKeys.java
```diff
@@ -140,6 +142,8 @@ public void handleInputEvent(InputEvent event) {
         } else if (keyPressed == KEY.ACTION2) {
             FactionPlayerHandler factionHandler = FactionPlayerHandler.get(Minecraft.getMinecraft().player);
             toggleBoundAction(factionHandler.getCurrentFactionPlayer(), factionHandler.getBoundAction2());
+        } else if (!isKeyDown(getKeyCode(KEY.SUCK))) {
```
maxanier: This event is trigger for any kind of input even a little mouse movement, which leads to dozens of EndSuckBlood messsages being sent per second. 
I would suggest setting a simple flag if the key is pressed (and the suck blood packet is sent) and then checking/clearing that flag here

maxanier (changes requested):
Functionality looks mostly good now.
Tested multiplayer as well as player feeding from player. All good.

Still some internal things to change

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: yeah the render stuff is killing me.  I looked into a potion effect but I wasn't sure how I could do client side rendering stuff on a potion effect.

src/main/java/de/teamlapen/vampirism/proxy/IProxy.java
tsheinen: no reason, I probably added it when I made the function signature and forgot to remove it when I didn't end up using it.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
maxanier: it is not that important, can always add it later if we want

src/main/java/de/teamlapen/vampirism/proxy/IProxy.java
maxanier: ok

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: fixed as of latest commit.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: fixed as of latest commit.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: fixed as of latest commit.  I removed the four times a second bite particle and added the altar of inspiration blood particles each bite.  The end of the particles is delayed enough that its pretty constant in between bites.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: yeah I commented it because when I made that change something was broken and I never uncommented it.  its fixed now

src/main/java/de/teamlapen/vampirism/client/core/ModKeys.java
tsheinen: I added a variable to store if the key is held down or not.  It'll now only trigger if the suck key is being released.

tsheinen:
Okay, I think I got everything you mentioned in the last commit.  The remaining three outstanding issues are the rendering, the NPC feeding, and crashing while feeding makes you stuck slow unless you feed again.  

As you mentioned, I think the ideal method of handling the rendering is by attaching it to a potion effect.  I looked through the Potion class and I couldn't find a good way of rendering client side stuff attached to a potion effect.  The few methods I did see all seemed to be inventory only.  

NPC feeding seems to be completely divorced from the player feeding method and would have to be individually implemented.  The main problem is that I have no idea how i'd need to change the AI to be compatible with a continuous feeding process.  

As for the crash while feeding error.  Basically if the feeding gets interrupted without getting a chance to end it sanely, the player gets stuck slow.  Assuming the server gets a chance to send log off events, it'll end the bite process in the logoff event.  As near as I can tell they don't keep the potion effect but only the attribute change.  This can be fixed by starting the bite process.  I've absolutely no clue how to fix this and why feeding does.  I've tried clearing all potion effects on logon, milk, adding a tick of the potion effect on logon and nothing seems to help.

maxanier:
Seems like the Potion method I thought about does no exist. Rendering would have to be done in VampirismHUDOverlay then, but I can implement that at a later point.

The feeding process of NPC is indeed mostly separated from the player feeding process (in VampireAIBiteNearbyEntity) and it already emulates a durational feeding process by looking at the victim for some time before biting. The only overlap is the EntityCreature#onBite method which is changed in this PR. Therefore vampire NPCs receive less blood. But I guess that is fine for now since it does not have that big an impact.

How can I reproduce that crash while feeding error? Maybe I can figure something out.
In theory the attribute change should be removed if the potion times out or is removed using EntityLivingBase#removePotionEffect.

Two more things I noticed:
- The client is not informed when the bite process stops (so it keeps sending stop bite packages)
- The bite counter keeps counting even if there is no victim and it is not reset when the process starts/stops

Moreover I currently need to press the bite key twice to start the biting, but I assume that is caused by one of the points above.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -216,15 +245,54 @@ public void biteEntity(int entityId) {
...
+            endBiting();
+            return;
+        }
+        PotionEffect effect = new PotionEffect(PotionBloodLoss.POTION, 20, 7);
+        ((EntityLivingBase) e).addPotionEffect(effect);
+
+        PotionEffect feedingEffect = new PotionEffect(PotionFeeding.POTION, 25);
+        player.addPotionEffect(feedingEffect);
+
+        VampLib.proxy.getParticleHandler().spawnParticles(player.world, ModParticles.FLYING_BLOOD_ENTITY, e.posX + 0.5, e.posY + 0.5, e.posZ + 0.5, 40, 0.1F, player.getRNG(), player, false);
```
maxanier: If you change the last flag to true, the particles will directly move towards the player and not explode all around. Furthermore you could reduce the number of particles from 40 to 10 or so

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -216,15 +245,54 @@ public void biteEntity(int entityId) {
...
         }
     }
 
+    public void biteVictim() {
+        EntityLivingBase e = ((EntityLivingBase) player.world.getEntityByID(victim));
+        if (e == null) {
+            endBiting();
+            return;
+        }
+        PotionEffect effect = new PotionEffect(PotionBloodLoss.POTION, 20, 7);
```
maxanier: It might look less chaotic if both potion effects do not cause particle effects as they tend to clutter the screen. If you the PotionEffect(Potion,duration,amplifier,ambient,particle) constructor you can disable them

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -216,15 +245,54 @@ public void biteEntity(int entityId) {
...
+                }
             } else {
                 VampirismMod.log.w(TAG, "Entity sent by client is not in reach " + entityId);
             }
         }
     }
 
+    public void biteVictim() {
+        EntityLivingBase e = ((EntityLivingBase) player.world.getEntityByID(victim));
+        if (e == null) {
```
maxanier: Currently the bite process continues for 20 ticks after the victim has died since it is only removed from the world after the death animation has finished. Maybe we can also check if the entity health is >0 or if it still has blood her.

tsheinen:
You can reproduce the feeding/effect error by feeding on an entity of some sort, pausing, and then forcibly ending the game process somehow.  

I'll see about changing the onBite calculations depending on if the biting entity is a player or not.  I also just noticed that double click issue, I've no idea where that came from.  I've noticed that it doesn't apply on the first bite of the world load.  I'll see if I can figure out whats causing that and fix it.

tsheinen:
I just (probably, its difficult to test) made the bite damage go back to the original if its an npc vampire.  I also made it sync the victim down when the bite ends and set the bite timer to 0 when biteEntity fires.  

I also fixed what was arguably the weirdest bug of my life.  So i'm sitting here watching via println the bitetimer stay at 0/1 during the feeding process and nothing happens.  I pause and set a breakpoint to step through it and see what was breaking it.... and it works fine.  After some testing I determined that it broke the feeding process only until the game was paused, after which it worked fine.  It should all now be fixed though.

maxanier:
Yeah, debugging MC is always fun :D 
Does this include the crash attribute bug? I was not able to reproduce it by the way

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -216,15 +245,59 @@ public void biteEntity(int entityId) {
...
+        PotionEffect effect = new PotionEffect(PotionBloodLoss.POTION, 20, 7, false, false);
+        ((EntityLivingBase) e).addPotionEffect(effect);
+
+        PotionEffect feedingEffect = new PotionEffect(PotionFeeding.POTION, 25, 1, false, false);
+        player.addPotionEffect(feedingEffect);
+
+        VampLib.proxy.getParticleHandler().spawnParticles(player.world, ModParticles.FLYING_BLOOD_ENTITY, e.posX + 0.5, e.posY + 0.5, e.posZ + 0.5, 10, 0.1F, player.getRNG(), player, true);
+
+        biteEntity(e);
+        if (!(e.getDistance(player) <= player.getEntityAttribute(EntityPlayer.REACH_DISTANCE).getAttributeValue() + 1) || e.isDead)
```
maxanier: Did some testing and turns out my last suggestion did not help the "delayed feeding stop", we also need to check the health here.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -462,6 +535,7 @@ public int onBite(IVampire biter) {
         int sucked = (int) (amt * perc);
         bloodStats.removeBlood(sucked, true);
         sync(this.bloodStats.writeUpdate(new NBTTagCompound()), true);
+        feedCounter = 30;
```
maxanier: Is this going to be used somewhere?

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -447,7 +520,7 @@ public void loadData(NBTTagCompound nbt) {
 
     @Override
     public int onBite(IVampire biter) {
-        float perc = biter instanceof IVampirePlayer ? 1F : 0.4F;
+        float perc = biter instanceof IVampirePlayer ? 0.2F : 0.08F;
```
maxanier, line 435: This issue was present before, but it has a bigger impact with the changed percentage:

If a player has a very low blood/food bar the onBite method will calculate a value of 0 even if there is blood/food left.
So we need to ensure we suck at least 1 blood, but only if there is actually 1 blood/food left

maxanier:
System looks good to me now, but we need to integrate the code a little better and remove the old unnecessary one (mainly the biteCooldown timer.

Currently most of the important checks are done in `VampirePlayer#biteEntity(EntityLivingBase)` which is only called once the bite timer ticks. I think we should move the `determineBiteType` check to the `VampirePlayer#biteEntity(int)` method which is called at first. 

The remaining code of `biteEntity(EntityLivingBase)` should be split up in something like `biteFeed(EntityLivingBase)` which handles the blood transfer part and is called why the feeding system and a `biteAttack(EntityLivingBase)` which handles the attack part and is directly called from `biteEntity(int)`.

Hope I managed to explain this properly

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: oh I added it as part of the overlay code and forgot to remove it.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: I think this is now resolved, as of https://github.com/TeamLapen/Vampirism/pull/355/commits/244c30b06a30c02c65cf52341bfa55f745e15cb0

tsheinen:
I split the bite code.  I definitely agree that makes it cleaner.  I'm less confident about the fix for the issue with integer division and blood calculation.  I use math.ceil while calculating the blood to ensure that making it an int rounds up instead.  I think it fixed it, but i'm not entirely sure I understand the issue.  

The crash attribute bug does not appear to be fixed.  To reproduce it I
* start the game
* start feeding on an entity by holding down V
* pause the game while still holding down V
* exit the game forcibly, I use the stop button in intellij
* load the game again

If you still can't reproduce it it's probably a lot less of a problem since it shouldn't come up often anyways.  It'd really only happen if the server crashed while someone was feeding.

src/main/java/de/teamlapen/vampirism/client/core/ModKeys.java
```diff
@@ -140,6 +145,9 @@ public void handleInputEvent(InputEvent event) {
         } else if (keyPressed == KEY.ACTION2) {
             FactionPlayerHandler factionHandler = FactionPlayerHandler.get(Minecraft.getMinecraft().player);
             toggleBoundAction(factionHandler.getCurrentFactionPlayer(), factionHandler.getBoundAction2());
+        } else if (SUCK_KEY_DOWN == true && !isKeyDown(getKeyCode(KEY.SUCK))) {
```
maxanier: Make this separate outside the else if contruct

src/main/java/de/teamlapen/vampirism/client/core/ModKeys.java
```diff
@@ -112,12 +114,15 @@ public void handleInputEvent(InputEvent event) {
         // get value here!
         if (keyPressed == KEY.SUCK) {
             RayTraceResult mouseOver = Minecraft.getMinecraft().objectMouseOver;
```
maxanier: Also check if `!SUCK_KEY_DOWN`

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -690,8 +695,14 @@ public void onUpdate() {
                 if (sync) {
                     sync(syncPacket, syncToAll);
                 }
-            } else {
 
+                if (victim != -1 && biteTickCounter >= 20) {
```
maxanier: Just a tiny improvement, but we only need to increase the biteTickCounter if there is a victim.
e.g. with something like
`if(victim !=-1 && biteTickCounter++ >= 20){`

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -708,6 +719,16 @@ public void onUpdate() {
                 ticksInSun = 0;
             }
 
+            if (biteTickCounter >= 5 && victim != -1) {
```
maxanier: Same here (need to swap order though)

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1065,69 +1102,136 @@ private void biteBlock(@Nonnull BlockPos pos, @Nonnull IBlockState blockState, @
...
+     */
+    public void biteEntity(int entityId) {
+        Entity e = player.getEntityWorld().getEntityByID(entityId);
+        if (player.isSpectator()) {
+            VampirismMod.log.w(TAG, "Player can't bite in spectator mode");
+            return;
+        }
+        if (e != null && e instanceof EntityLivingBase) {
+            if (e.getDistance(player) <= player.getEntityAttribute(EntityPlayer.REACH_DISTANCE).getAttributeValue() + 1) {
+                if (determineBiteType(((EntityLivingBase) e)) == BITE_TYPE.ATTACK || victim_bite_type == BITE_TYPE.ATTACK_HUNTER) {
```
maxanier: I think the second check (==ATTACK_HUNTER) does not respect the new value, but a older value previously stored in victim_bite_type.

Also bite type is determined up to two times in the whole method. Probably better to get it only once in the beginning (Store it either directly in victim_bite_type or in some intermediate variable)

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1065,69 +1102,136 @@ private void biteBlock(@Nonnull BlockPos pos, @Nonnull IBlockState blockState, @
...
+                    biteAttack((EntityLivingBase) e, victim_bite_type == BITE_TYPE.ATTACK_HUNTER);
+                } else {
```
maxanier: If bite type == NONE, we do not want to start feeding

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1065,69 +1102,136 @@ private void biteBlock(@Nonnull BlockPos pos, @Nonnull IBlockState blockState, @
...
-            } else if (type == BITE_TYPE.ATTACK_HUNTER) {
-                if (entity instanceof EntityPlayer && ItemHunterCoat.isFullyEquipped((EntityPlayer) entity)) {
-                    player.attackEntityFrom(DamageSource.causeThornsDamage(entity), damage);
-                }
-            }
-            if (specialAttributes.poisonous_bite) {
-                entity.addPotionEffect(new PotionEffect(MobEffects.POISON, (int) (Balance.vps.POISONOUS_BITE_DURATION * 20 * (getSpecialAttributes().bat ? 0.2F : 1F)), 1));
-            }
-        } else if (type == BITE_TYPE.NONE) {
+        } else if (victim_bite_type == BITE_TYPE.NONE) {
```
maxanier: Should be handled in biteEntity(int)

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1065,69 +1102,136 @@ private void biteBlock(@Nonnull BlockPos pos, @Nonnull IBlockState blockState, @
...
-        biteCooldown = Balance.vp.BITE_COOLDOWN;
         if (blood > 0) {
             drinkBlood(blood, saturationMod);
             //TODO player.addStat(Achievements.suckingBlood, 1);
             NBTTagCompound updatePacket = bloodStats.writeUpdate(new NBTTagCompound());
             updatePacket.setInteger(KEY_SPAWN_BITE_PARTICLE, entity.getEntityId());
             sync(updatePacket, true);
         }
-        if (type == BITE_TYPE.SUCK_BLOOD || type == BITE_TYPE.SUCK_BLOOD_CREATURE || type == BITE_TYPE.SUCK_BLOOD_PLAYER) {
+        if (victim_bite_type == BITE_TYPE.SUCK_BLOOD || victim_bite_type == BITE_TYPE.SUCK_BLOOD_CREATURE || victim_bite_type == BITE_TYPE.SUCK_BLOOD_PLAYER) {
```
maxanier: I don't know why I but the advancement trigger here before, I think it would be best to move it to the end of the `if(blood > 0)` condition, so we don't need to check the bite type again.

maxanier (changes requested):

maxanier:
`Math.ceil` should work fine I think :+1: 

I still cannot reproduce this issue, so I think it is fine. Even if it should occur the attribute change can be fixed easily.

I noticed a weird bug (I think) in MC or Forge, I need to investigate further.
When checking if a KeyBinding is pressed via Keybinding#isPressed , it is supposed to only return true once. However, right after joining a world it keeps returning true when holding the key down. Only after opening and closing some GUI (e.g. chat) it works as intended.

Therefore, the game is currently sending dozens of SUCK_BLOOD packages while feeding (if no GUI has been opened before). As a temporary fix until the underlying issue has been fixed I would suggest
using the SUCK_KEY_DOWN flag which is present anyway to only send trigger the feeding process if the key has not been down before.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
tsheinen: i've no idea what on earth I was thinking here, its fixed locally and i'll commit once i've fixed the rest of this stuff

tsheinen:
It's really strange you can't reproduce it.  I wonder whats causing it?  Oh, well, I suppose I'll look into it in more depth if anyone else has issues with it.  

I've updated and fixed all the stuff you mentioned.

src/main/java/de/teamlapen/vampirism/client/core/ModKeys.java
```diff
@@ -53,6 +53,8 @@
     private static KeyBinding ACTION1 = new KeyBinding(ACTIVATE_ACTION1, KeyConflictContext.IN_GAME, KeyModifier.ALT, Keyboard.KEY_1, CATEGORY);
     private static KeyBinding ACTION2 = new KeyBinding(ACTIVATE_ACTION2, KeyConflictContext.IN_GAME, KeyModifier.ALT, Keyboard.KEY_2, CATEGORY);
 
+    private boolean SUCK_KEY_DOWN = false;
```
maxanier: Probably better if lowercase since not a constant

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -48,27 +62,37 @@
...
 import net.minecraftforge.fluids.capability.IFluidHandler;
 import net.minecraftforge.fml.common.FMLCommonHandler;
 import net.minecraftforge.fml.common.Loader;
+import net.minecraftforge.fml.common.eventhandler.EventPriority;
+import net.minecraftforge.fml.common.eventhandler.SubscribeEvent;
 import net.minecraftforge.fml.common.gameevent.TickEvent;
 import net.minecraftforge.fml.relauncher.ReflectionHelper;
+import net.minecraftforge.fml.relauncher.Side;
+import net.minecraftforge.fml.relauncher.SideOnly;
+import org.lwjgl.opengl.GL11;
```
maxanier: Unused imports should be removed. I usually tick the box "optimise imports" box in the intellij commit dialog. it should only affect VCS changed code

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -144,6 +168,10 @@ public NBTTagCompound serializeNBT() {
     private IVampireVision activatedVision = null;
     private Method reflectionMethodSetSize = null;
 
+    private int victim = -1;
```
maxanier: probably best to prefix this variable with `feed_` or `bite_` to make it more distinct and adjust the other ones accordingly. VampirePlayer is quite a big class

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1065,69 +1099,134 @@ private void biteBlock(@Nonnull BlockPos pos, @Nonnull IBlockState blockState, @
...
+                VampirismMod.log.w(TAG, "Entity sent by client is not in reach " + entityId);
+            }
+        }
+    }
+
+
+    /**
+     * This is called every 20 ticks in onUpdate() to run the continuous feeding effect
+     */
+    public void biteVictim() {
```
maxanier: Should be private I think.
We could also consider renaming this to something like `updateFeeding` and the endBite to `endFeeding` or `stopFeeding`.  Thereby it is more recognizable as a method which is called regularly and clearly separated from the biteAttack / biteFeed.

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1065,69 +1099,134 @@ private void biteBlock(@Nonnull BlockPos pos, @Nonnull IBlockState blockState, @
...
+        EntityLivingBase e = ((EntityLivingBase) player.world.getEntityByID(victim));
```
maxanier: It is very unlikely, though still possible that getEntityByID returns an `Entity` instead of the expected `EntityLivingBase`.
In modded MC it is usually better to be safe than sorry :D

src/main/java/de/teamlapen/vampirism/proxy/ClientProxy.java
```diff
@@ -75,6 +77,12 @@ public void renderScreenFullColor(int ticksOn, int ticksOff, int color) {
         if (overlay != null) overlay.makeRenderFullColor(ticksOn, ticksOff, color);
     }
 
+    @Nullable
+    @Override
+    public Entity getMouseOverEntity() {
+        return Minecraft.getMinecraft().objectMouseOver.entityHit;
```
maxanier: Same here, in theory `objectMouseOver` should never be null, but even Vanilla does null check it every time, so we probably should do that too

src/main/java/de/teamlapen/vampirism/potion/PotionBloodLoss.java
```diff
@@ -0,0 +1,17 @@
+package de.teamlapen.vampirism.potion;
+
+
+import net.minecraft.entity.SharedMonsterAttributes;
+
+
+
+public class PotionBloodLoss extends VampirismPotion {
+    public static PotionFeeding POTION= new PotionFeeding("Blood Loss", false, Integer.parseInt("FF0000", 16));
```
maxanier: I am surprised that it is working :D
Potions should be registered to Forge and their names are supposed to be lowercase and one word. 
However, it looks like it is not enforces right now. Still probably better to be prepared for future changes.

The other Potions are instantiated in `ModPotions#registerPotions`. If you add a `public static final Potion name = getNull();` at the top of the ModPotions class with exactly the same name as the potion name. It will be automatically filled with the instance by Forge magic.

maxanier (changes requested):
Perfect. So one final round of style changes to make it more aligned with the rest of the code and we are good to go :)

src/main/java/de/teamlapen/vampirism/potion/PotionBloodLoss.java
tsheinen: tbh i'm not sure why either of those still exist.  They're both literally reskinned slowness.  I think I was intending to tie the overlay to it?  I might as well just use the normal slowness effect.

maxanier (approved):

maxanier:
Perfect. Thank you very much for your contribution, I think this feature will make the feeding more immersive. I will release new alpha build later today so players can try it.

Edit:
Only after merging I found two minor issues so I quickly fixed them in b25d39ca5ee60fad29b261eba0a055c11e4c8231

Issue #345, opened 2018-08-15
Make vampire NPCs stronger


maxanier:
The vampire mods are quite common but rather weak. Maybe it would be for "lore" friendly if they would spawn more rarely and be stronger.
The low level vampires should not be too strong, so level 0 players are not in danger too much, but we already have a leveling system for the basic vampires in place anyway so maybe the higher level ones could get a boost.
In addition it would be awesome if they could use some of the skills the player has on hand #97

**Edit**: Cheaterpaul is currently working on this here: https://github.com/Cheaterpaul/Vampirism/tree/vampire_npc

Cheaterpaul:
do you have a link for the leveling system of the basic vampire?

maxanier:
EntityBasicVampire implements https://github.com/TeamLapen/Vampirism/blob/1.12/src/api/java/de/teamlapen/vampirism/api/difficulty/IAdjustableLevel.java

On spawn (somewhere in ModEntityEventhandler I guess (maybe in the lib package)) the level is decided and set and this is called https://github.com/TeamLapen/Vampirism/blob/6679cc77af31d60b3779102effa804c54da1d508/src/main/java/de/teamlapen/vampirism/entity/vampire/EntityBasicVampire.java#L97

This also updates the entity attributes which vary depending on the level 
https://github.com/TeamLapen/Vampirism/blob/6679cc77af31d60b3779102effa804c54da1d508/src/main/java/de/teamlapen/vampirism/entity/vampire/EntityBasicVampire.java#L271

Cheaterpaul:
maybe growing level stats, like the e funktion?

https://github.com/TeamLapen/Vampirism/blob/6679cc77af31d60b3779102effa804c54da1d508/src/main/java/de/teamlapen/vampirism/entity/vampire/EntityBasicVampire.java#L43

could me more?

maxanier:
I would prefer to not increase their basic attributes too much. It is easy, but a little bit boring. A lot cooler (but also more time consuming) would be to make them use some vampire skills/actions.

Nonetheless we can increase their stats a little, to make them a little bit stronger than Zombies at least.

shawndream:
I was going to comment here that rather than just making them stronger they should be able to summon minions behind you when angered.  (Then I found my first Vampire Baron and saw you've got that in there-ish).

[...]

Issue #97, opened 2016-07-06
NPC should use skills/actions


maxanier:
Normal vampires/hunters only simple ones, advanced ones more special ones. 

Hunters should use crossbows 
E. G. Batmode, freeze, sleep in coffin,

maxanier:
Closed with #378 
@Cheaterpaul BTW, do you have permission to close issues and edit projects?

PR #378, opened 2019-01-15
NPC actions

Cheaterpaul:
#97 
#345


src/api/java/de/teamlapen/vampirism/api/entity/actions/IEntityAction.java
```diff
@@ -0,0 +1,10 @@
+package de.teamlapen.vampirism.api.entity.actions;
+
+import net.minecraftforge.registries.IForgeRegistryEntry;
+
+public interface IEntityAction extends IForgeRegistryEntry<IEntityAction> {
+
+    int getDuration(int level);
```
maxanier: Why do we have a getDuration here? Instant actions should not need one, do they?

src/api/java/de/teamlapen/vampirism/api/entity/actions/ILastingAction.java
```diff
@@ -0,0 +1,16 @@
+package de.teamlapen.vampirism.api.entity.actions;
+
+import de.teamlapen.vampirism.api.entity.factions.IFactionEntity;
+
+public interface ILastingAction<T extends IFactionEntity> extends IEntityAction {
+
+    int getDuration(int level);
+
+    int getCooldown(int level);
```
maxanier: Why is this here again? Already defined in IEntityAction

src/api/java/de/teamlapen/vampirism/api/entity/actions/ILastingAction.java
```diff
@@ -0,0 +1,16 @@
+package de.teamlapen.vampirism.api.entity.actions;
+
+import de.teamlapen.vampirism.api.entity.factions.IFactionEntity;
+
+public interface ILastingAction<T extends IFactionEntity> extends IEntityAction {
```
maxanier: Maybe add Javadoc to the methods explaining when they are called (and especially if they are called client, server side only or both)

src/main/java/de/teamlapen/vampirism/client/render/particle/HealingParticle.java
```diff
@@ -0,0 +1,83 @@
...
+        this.move(this.motionX, this.motionY, this.motionZ);
+        this.motionX *= 0.9800000190734863D;
+        this.motionY *= 0.9800000190734863D;
+        this.motionZ *= 0.9800000190734863D;
+
+    }
+
+    @Override
+    public void renderParticle(BufferBuilder buffer, Entity entityIn, float partialTicks, float rotationX, float rotationZ, float rotationYZ, float rotationXY, float rotationXZ) {
+        int i = (int) (((float) this.particleAge + partialTicks) * 3.0F / (float) this.particleMaxAge);
```
maxanier: What does this do?
particleAge < particleMaxAge
0<partialTicks<3

-> i <=3

src/main/java/de/teamlapen/vampirism/core/ModParticles.java
```diff
@@ -1,33 +1,36 @@
...
     public static final ResourceLocation FLYING_BLOOD_ENTITY = new ResourceLocation(REFERENCE.MODID, "flying_blood_entity");
     public static final ResourceLocation HALLOWEEN = new ResourceLocation(REFERENCE.MODID, "halloween");
+    public static final ResourceLocation HEAL = new ResourceLocation(REFERENCE.MODID, "heal");
 
     /**
      * Arguments: Particle ID (Vanilla texture,int), TicksToLive(int), Color(int), [speed modifier (double)]
      */
     public static final ResourceLocation GENERIC_PARTICLE = new ResourceLocation(REFERENCE.MODID, "generic");
 
+    public static TextureAtlasSprite modParticleAtlas;
```
maxanier: Probably do not need this anymore

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionManager.java
```diff
@@ -0,0 +1,23 @@
...
+import de.teamlapen.vampirism.core.VampirismRegistries;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.registries.IForgeRegistry;
+import java.util.List;
+
+public class EntityActionManager implements IEntityActionManager {
+
+    @Override
+    public List<IEntityAction> getAllEntityActions() {
+        return Lists.newArrayList(VampirismRegistries.ENTITYACTIONS.getValues());
```
maxanier: .getValues() is deprecated use .getValuesCollection() (same result)

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,106 @@
...
+        } else if (action instanceof IInstantAction) {
+            ((IInstantAction) action).activate(entity);
+            this.action = null;
+        }
+    }
+
+    private void newAction() {
+        if (rand.nextInt(1) == 0) { // TODO usable balance
+            try {
+                action = availableActions.get(rand.nextInt(availableActions.size()));
```
maxanier: If availableAction is empty, this throws OutOfBounds (nextInt(0))

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,106 @@
...
+                System.out.println(action.getClass().getName()); // TODO remove
+            } catch (NullPointerException e) {
```
maxanier: There should not be a NullPointerException possible here

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,106 @@
...
+        this.entity = entityIn;
+        this.availableActions = actions;
+    }
+
+    @Override
+    public boolean shouldExecute() {
+        if (entity.getAttackTarget() instanceof EntityPlayer) {
+            return true;
+        }
+        return false;
```
maxanier: Maybe check if availableActions is not empty

src/main/java/de/teamlapen/vampirism/core/ModParticles.java
```diff
@@ -157,5 +160,24 @@ public NBTTagCompound createParticleInfo(Object... param) {
...
+
+            @Override
+            public NBTTagCompound createParticleInfo(Object... param) {
+                return new NBTTagCompound();
+            }
+
+            @SideOnly(Side.CLIENT)
+            @Override
+            public Particle createParticle(World world, double posX, double posY, double posZ, Object... param) {
+                return new HealingParticle(world, posX, posY, posZ, posZ, posZ, posZ);
```
maxanier: Why are you passing posZ for all movement arguments?

Particles look cool, but they still move in a unnatural way, diagonally up? maybe add some more randomness, and make them slower

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -48,6 +54,11 @@
     public EntityVampirism(World world) {
         super(world);
         moveTowardsRestriction = new EntityAIMoveTowardsRestriction(this, 1.0F);
+        this.actionTasks = new EntityAITasks(world != null && world.profiler != null ? world.profiler : null);
```
maxanier: Do we really need our own EntityAITasks. Since there is only a single actual AITask ever added to this list, we could either add the task to the normal `tasks` list or just directly perform the task update in the `onUpdate` method

src/main/java/de/teamlapen/vampirism/entity/action/actions/InvisibleEntityAction.java
```diff
@@ -0,0 +1,45 @@
...
+        if (!entity.getRepresentingEntity().isInvisible()) {
+            entity.getRepresentingEntity().setInvisible(true);
+        }
+        return false;
+    }
+
+    @Override
+    public void activate(T entity, int duration) {
+        for (int i = 0; i < 3; i++) {
+            Minecraft.getMinecraft().world.spawnParticle(EnumParticleTypes.EXPLOSION_LARGE, entity.posX, entity.posY, entity.posZ, 0, 0.5, 0);
```
maxanier: Explosion? :D

src/main/java/de/teamlapen/vampirism/entity/action/actions/SpeedEntityAction.java
```diff
@@ -0,0 +1,55 @@
...
+        return true;
+    }
+
+    @Override
+    public void activate(T entity, int duration) {
+    }
+
+    public void effect(T entity, int duration) {
+        if (duration % 5 == 0) {
+            Minecraft.getMinecraft().world.spawnParticle(EnumParticleTypes.CLOUD, entity.posX, entity.posY, entity.posZ, -entity.motionX, 0.05, -entity.motionZ);
```
maxanier: Do they actually spawn? I think effect is called server side and this spawnParticle method only works client side.
Also we probably do not need a dedicated effect method, this could be moved into onUpdate

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,106 @@
...
+/*
+ * Usage every {@link IFactionEntity} like Hunter/Vampire entities
+ */
+public class EntityAIUseAction<T extends EntityVampirism & IFactionEntity & IAdjustableLevel> extends EntityAIBase {
+
+    private T entity;
+    private List<DefaultEntityAction> availableActions;
+    private int cooldown = 0;
+    private int duration = 0;
+    private DefaultEntityAction action;
```
maxanier: Is it possible to use IEntityAction instead of DefaultEntityAction here?

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,106 @@
...
+    public void startExecuting() {
+        action = null;
+        cooldown = 100;
+    }
+
+
+    @Override
+    public void updateTask() {
+        if (duration > 0) {
+            duration--;
```
maxanier: I am a little bit confused about the duration/cooldown handling in this class.
Duration can become negative, duration is set even though a instant action is selected, duration is not reset so updateAction might be called even if action==null.

This probably needs another look

maxanier (changes requested):
Thank you for working on this.

These are some quite cool features/additions. However, there are a few things that I do not fully understand or need to be changed.

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
Cheaterpaul: Adding EntityAITask to an existing task list will interrupt the task (possibly with Write to Nbt). The onUpdate method should be possible if you prefer.

src/main/java/de/teamlapen/vampirism/entity/action/actions/SpeedEntityAction.java
Cheaterpaul: todo: check with server

src/main/java/de/teamlapen/vampirism/client/render/particle/HealingParticle.java
```diff
@@ -0,0 +1,83 @@
...
+        this.motionY -= 0.04D * (double) this.particleGravity;
+        this.move(this.motionX, this.motionY, this.motionZ);
+        this.motionX *= 0.9800000190734863D;
+        this.motionY *= 0.9800000190734863D;
+        this.motionZ *= 0.9800000190734863D;
+
+    }
+
+    @Override
+    public void renderParticle(BufferBuilder buffer, Entity entityIn, float partialTicks, float rotationX, float rotationZ, float rotationYZ, float rotationXY, float rotationXZ) {
```
Cheaterpaul, line 33: render the particle pointing towards the player

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
Cheaterpaul: check if its enough

src/main/java/de/teamlapen/vampirism/entity/action/actions/InvisibleEntityAction.java
Cheaterpaul: now explosion_normal?  if its ok or changing to another or custom (recommendation) or is explosion_large ok

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
Cheaterpaul: IllegalArgumentException (0 is not positive), but checked in shouldExecute()

src/api/java/de/teamlapen/vampirism/api/entity/actions/ILastingAction.java
Cheaterpaul: check if its enough

src/main/java/de/teamlapen/vampirism/entity/action/actions/SpeedEntityAction.java
Cheaterpaul: test this by joining with client to server?
return Invalid session

Cheaterpaul: how to call onUpdate() in SpeedEntityAction or activate() in InvisibilEntityAction client side?

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
Cheaterpaul: Implemented in shouldExecute()

src/main/java/de/teamlapen/vampirism/core/ModParticles.java
```diff
@@ -176,5 +187,100 @@ public Particle createParticle(World world, double posX, double posY, double pos
                 return new HealingParticle(world, posX, posY, posZ);
             }
         });
+
+        ParticleHandler.registerParticle(CLOUD, new ParticleHandler.ICustomParticleFactory() {
```
Cheaterpaul, line 183: not nice but functional

src/main/java/de/teamlapen/vampirism/core/ModParticles.java
```diff
@@ -189,97 +186,33 @@ public Particle createParticle(World world, double posX, double posY, double pos
...
-                            }
-                        }
-
-                        if (this.onGround) {
-                            this.motionX *= 0.699999988079071D;
-                            this.motionZ *= 0.699999988079071D;
-                        }
-                    }
-                };
+                IParticleFactory par = new ParticleCloud.Factory();
```
maxanier: Is it possible to store the ParticleCloud factory as a  (final) variable inside this factory class/object?

src/main/java/de/teamlapen/vampirism/core/ModParticles.java
```diff
@@ -189,97 +186,33 @@ public Particle createParticle(World world, double posX, double posY, double pos
...
-                return nbt;
-            }
 
-            @Nonnull
             @SideOnly(Side.CLIENT)
+            @Nonnull
             @Override
             public Object[] readParticleInfo(NBTTagCompound nbt) {
-                Object[] data = new Object[5];
+                Object[] data = new Object[nbt.hasKey("3") ? 4 : 3];
```
maxanier: You are checking for a third (or rather fourth) value, but not loading it

src/main/java/de/teamlapen/vampirism/core/ModParticles.java
```diff
@@ -189,97 +186,33 @@ public Particle createParticle(World world, double posX, double posY, double pos
         });
 
         ParticleHandler.registerParticle(CLOUD, new ParticleHandler.ICustomParticleFactory() {
```
maxanier, line 183: Looks like a decent solution now :)

src/main/java/de/teamlapen/vampirism/entity/action/actions/SpeedEntityAction.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.entity.SharedMonsterAttributes;
+import net.minecraft.entity.ai.attributes.AttributeModifier;
+import net.minecraft.util.EnumParticleTypes;
+import net.minecraft.util.math.MathHelper;
+import java.util.UUID;
+import io.netty.util.internal.ThreadLocalRandom;
+
+public class SpeedEntityAction<T extends EntityVampirism & IFactionEntity & IAdjustableLevel> extends DefaultEntityAction implements ILastingAction<T> {
+
+    private UUID uuid = MathHelper.getRandomUUID(ThreadLocalRandom.current());
```
maxanier: Use a constant value here, this should not change over time.
You can generate one e.g. at https://www.uuidgenerator.net/

maxanier (changes requested):

Cheaterpaul:
should the actions be more frequent and weaker or less frequent and stronger?

maxanier:
Not sure, I have asked for feedback on Discord but did not get any response yet.
Personally, I would say, not too often. Basic vampires should not use it more than once during a fight on average.

Cheaterpaul:
should now be complete, with the exception of balance and actions used by entities

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -33,6 +36,8 @@
     private final EntityAIBase moveTowardsRestriction;
     protected boolean hasArms = true;
     protected boolean peaceful = false;
+    /** available actions for AI task */
+    public EntityAIUseAction<?> entityAIUseAction;
```
maxanier: Should be protected probably. We don't want anyone messing with this

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,149 @@
...
+        if (action instanceof ILastingAction) {
+            ((ILastingAction<T>) action).onUpdate(entity, duration);
+        }
+    }
+
+    /**
+     * sets a new random action from {@link availableActions} for the entity,
+     * otherwise a placeholder
+     */
+    private void newAction() {
```
maxanier: I think it would be cleaner if this would just return an action which would then be set as active.
Maybe `selectNewAction` or `getRandomAction` would be a more meaningful name then.

Also, the dummy action seems unessesarry, maybe just return null here, and then check in updateTasks for null (and set a cooldown then)

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,149 @@
...
+                }
+            };
+        }
+    }
+
+    /**
+     * use this method if EntityAIUseAction is used in onUpdate() in an Entity
+     */
+    public void handle() {
+        if (shouldExecute()) {
```
maxanier: Not sure, but if the attack target changes to null or a non player, this stops updating the task, but does not reset/deactivate it?

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,149 @@
...
+        } else if (!availableActions.isEmpty()) {
+            if (flag) {
+                resetTask();
+            }
+            flag = false;
+        }
+    }
+
+    public void reset() {
+        for (IEntityAction action : availableActions) {
```
maxanier: Why are you resetting all available actions? There should be only one active at most.

src/main/java/de/teamlapen/vampirism/entity/ai/EntityAIUseAction.java
```diff
@@ -0,0 +1,149 @@
...
+import net.minecraft.entity.ai.EntityAIBase;
+import net.minecraft.entity.player.EntityPlayer;
+import java.util.List;
+import java.util.Random;
+
+/**
+ * Usage for every {@link IFactionEntity} like Hunter/Vampire entities
+ * can be used as AITask in custom task list or with {@link handle()} in UpdateLiving in an Entity
+ */
+public class EntityAIUseAction<T extends EntityVampirism & IFactionEntity & IAdjustableLevel> extends EntityAIBase {
```
maxanier: Currently only EntityBasicVampire makes use of this, but not in the AITask, but with the handle() method.
Are you actually planning to use it in the AITask for certain entities?

If not, you could consider migrating away from the AI concept and do something similar like the ActionHandler and IActionPlayer for the entity action.

maxanier (changes requested):
Looks almost ready now :)

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -168,6 +178,9 @@ public void readEntityFromNBT(NBTTagCompound nbt) {
                 this.setMoveTowardsRestriction(nbt.getInteger("moveHomePrio"), true);
             }
         }
+        if (nbt.hasKey("activeAction")) {
```
maxanier: Can we move this load process into the Entity Action Handler (so just call `entityActionHandler.readFromNBT)`. Thereby it could more easily be used by other entity classes.

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -168,6 +178,9 @@ public void readEntityFromNBT(NBTTagCompound nbt) {
...
+            entityActionHandler.resetAction(EntityActionManager.actionsByID.get(nbt.getInteger("activeAction")));
```
maxanier: the map get could return null (if an action should be removed in a future version). Probably fine since resetAction accepts null values.

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -190,6 +203,11 @@ public void writeEntityToNBT(NBTTagCompound nbt) {
                 nbt.setInteger("homeMovePrio", moveTowardsRestrictionPrio);
             }
         }
+        if (entityActionHandler != null) {
```
maxanier, line 209: Same as for the load process: `actionHandler.writeToNBT()`

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -326,4 +344,18 @@ protected void updateAITasks() {
...
+    /**
+     * gets a list of action combos and adding a random combo to the available actions of the IFaction entity
+     * 
+     * @param actionstmp
+     *            list of action combos
+     */
+    protected void setAvailableActions(List<IEntityAction[]> actionstmp) {
+        if (actionstmp.size() == 0)
+            return;
+        entityActions = new ArrayList<IEntityAction>();
```
maxanier: Maybe assign a list instance on definition at the top of the class and just clear it here. Therefore, the list is never null. Shouldn't make a big difference though

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -326,4 +344,18 @@ protected void updateAITasks() {
...
+        for (IEntityAction e : actionstmp.get(this.getRNG().nextInt(actionstmp.size())))
```
maxanier: The code is not very straightforward here. Maybe you can untangle it a little bit and improve the documentation.
Maybe replacing the loop with an add all would be better: `entityActions.addAll(Arrays.asList(actionstmp.get(this.getRNG().nextInt(actionstmp.size()))));`

It adds all actions from one of the supplied arrays?
Currently all supplied arrays only contain one value

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,141 @@
...
+                }
+            }
+        }
+    }
+
+    public IEntityAction getAction() {
+        return action;
+    }
+
+    public boolean getFlag() {
```
maxanier: Not a very descriptive name :D

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionManager.java
```diff
@@ -0,0 +1,28 @@
...
+import de.teamlapen.vampirism.api.entity.actions.IEntityActionManager;
+import de.teamlapen.vampirism.core.VampirismRegistries;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.registries.IForgeRegistry;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class EntityActionManager implements IEntityActionManager {
+    public static int ID = 0;
```
maxanier: We shouldn't use public maps but rather getter and setter/adder methods, so we can more easily change things afterwards (and give this class more control like incrementing the id itself).

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionManager.java
```diff
@@ -0,0 +1,28 @@
...
+    public static final Map<Integer, IEntityAction> actionsByID = new HashMap<>();
```
maxanier: Why do you use an integer map. All actions should have a registered registry name available and should also be element of VampirismRegistries.ENTITYACTIONS)

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -33,6 +37,9 @@
     private final EntityAIBase moveTowardsRestriction;
     protected boolean hasArms = true;
     protected boolean peaceful = false;
+    /** available actions for AI task & task */
+    protected EntityActionHandler<?> entityActionHandler;
+    protected List<IEntityAction> entityActions;
```
maxanier: Maybe we can get rid of this variable now and pass the values directly to the action handler

maxanier (changes requested):
One step closer, but there are some more things to discuss

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
Cheaterpaul: actions used by classes are not finished

maxanier:
Just noticed that the automated builds are failing https://travis-ci.org/TeamLapen/Vampirism/builds/487692314

Anything inside the `api` package must not refer to anything outside of it. The api package might be shipped without the rest of Vampirism present.

Cheaterpaul:
I have a problem with the vampire level when it's set, everything is fine, but when I call getLevel (), -1 is returned.
https://github.com/Cheaterpaul/Vampirism/blob/eeda2f02805066c26af29cf899b229e60795e0c3/src/main/java/de/teamlapen/vampirism/entity/vampire/EntityBasicVampire.java#L96-L99

https://github.com/Cheaterpaul/Vampirism/blob/eeda2f02805066c26af29cf899b229e60795e0c3/src/main/java/de/teamlapen/vampirism/entity/vampire/EntityBasicVampire.java#L290

can't figure it out at the moment, if you have an idea.

maxanier:
It is only set after the entity has been spawned.
The `ModEntityEventHandler` (or similar) listens for EntitySpawn events and checks for `IAdjustableLevel` creatures. For such creatures it calculates a local difficulty based on the players around. Then it calls `setLevel`.

Maybe it is possible to add/modify the available actions in set level.

src/main/java/de/teamlapen/vampirism/entity/vampire/EntityVampireBase.java
```diff
@@ -145,7 +145,7 @@ public boolean isGettingSundamage(boolean forceRefresh) {
 
     @Override
     public boolean isIgnoringSundamage() {
-        return this.isPotionActive(ModPotions.sunscreen);
+        return this.isPotionActive(ModPotions.sunscreen) || this.ignoreSundamageAction;
```
maxanier: Is it possible to just use the potion for the entity action instead of having an additional attribute here?

src/main/java/de/teamlapen/vampirism/entity/vampire/EntityBasicVampire.java
```diff
@@ -282,35 +280,37 @@ protected void updateEntityAttributes() {
...
      */
-    protected List<IEntityAction> getAvailableActions(int level) {
-        List<IEntityAction[]> actionstmp = new ArrayList<IEntityAction[]>();
-        switch (level) {
-            case 0:
-                actionstmp.add(new IEntityAction[] { EntityActions.entity_regeneration });
-                actionstmp.add(new IEntityAction[] { EntityActions.entity_speed });
-                actionstmp.add(new IEntityAction[] { EntityActions.entity_bat_spawn, EntityActions.entity_regeneration });
-                actionstmp.add(new IEntityAction[] { EntityActions.entity_bat_spawn });
+    @Override
```
maxanier: It would be perfect if this could somehow be moved into the action registry and the actions themselves.
Maybe each action is registered with (or provides) the entity class(es) it fits as well as some kind of indication if it is rather a lower or a higher tier action. 
Then the code within the entity itself would only have to provide some sort of level (basic vampire: low to medium, advanced vampire: high, baron: ultimate) and it's class

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -348,11 +353,57 @@ protected void updateAITasks() {
...
+        AttributeModifier damageModifier;
+        AttributeModifier speedModifier;
+
+        EntityClass(double healthModifier, double damageModifier, double speedModifier) {
+            this.healthModifier = new AttributeModifier("entity_class_health", healthModifier, 1);
+            this.damageModifier = new AttributeModifier("entity_class_damage", damageModifier, 1);
+            this.speedModifier = new AttributeModifier("entity_class_speed", speedModifier, 1);
+        }
+
+        public static EntityClass getRandomClass() {
```
maxanier: Avoid recreating a new random every time. You could pass the entity's rng here

src/main/java/de/teamlapen/vampirism/entity/EntityVampirism.java
```diff
@@ -348,11 +353,57 @@ protected void updateAITasks() {
...
+    }
+
+    protected void setupEntityClass() {
+        entityclass = EntityClass.getRandomClass();
+        this.getEntityAttribute(SharedMonsterAttributes.MAX_HEALTH).applyModifier(entityclass.getHealthModifier());
+        this.getEntityAttribute(SharedMonsterAttributes.ATTACK_DAMAGE).applyModifier(entityclass.getDamageModifier());
+        this.getEntityAttribute(SharedMonsterAttributes.MOVEMENT_SPEED).applyModifier(entityclass.getSpeedModifier());
+    }
+
+    public enum EntityClass {
```
maxanier: Maybe rename this to EntityClassType or EntityActionType or EntityFightClass.
It took me a minute to find out that it is completely unrelated to Java classes.

Could be moved in some API class as well

src/main/java/de/teamlapen/vampirism/entity/action/vampire/SunscreamEntityAction.java
```diff
@@ -1,17 +1,19 @@
...
-import de.teamlapen.vampirism.api.entity.vampire.IVampire;
+import de.teamlapen.vampirism.api.entity.factions.IFactionEntity;
 import de.teamlapen.vampirism.config.Balance;
 import de.teamlapen.vampirism.core.ModPotions;
-import de.teamlapen.vampirism.entity.vampire.EntityVampireBase;
+import de.teamlapen.vampirism.entity.EntityVampirism;
 import net.minecraft.potion.PotionEffect;
 
-public class SunscreamEntityAction<T extends EntityVampireBase & IVampire & IAdjustableLevel> extends DefaultEntityAction implements ILastingAction<T> {
+public class SunscreamEntityAction<T extends EntityVampirism & IFactionEntity & IAdjustableLevel> extends VampireEntityAction implements ILastingAction<T> {
```
maxanier: Sunscreen :)

maxanier:
Looks very cool now :+1: 
Let me know if you think it is ready to merge and I will have a final look

Cheaterpaul:
ready for a final look

must remove attribute for cloudParticleFactory because IParticleFactory is ClientSide only

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+ * is used with {@link handle()} in UpdateLiving in an EntityVampirism
+ */
+public class EntityActionHandler<T extends EntityVampirism & IFactionEntity & IAdjustableLevel> {
+
+    private T entity;
+    private List<IEntityAction> availableActions;
+    private int preActivation = 0;
+    private int cooldown = 0;
+    private int duration = 0;
+    private float healthForDisruption;
```
maxanier: Maybe `healthTresholdForDisruption`

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+    public void deactivateAction() {
+        deactivateAction(null);
+    }
+
+    /**
+     * reset the given {@link IEntityAction} or if null, reset the {@link action}
+     * 
+     * @param actionIn
+     */
+    @Nullable
```
maxanier: Unnecessary

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+        updatePreAction(null);
+    }
+
+    /**
+     * updates the given {@link IEntityAction} pre activation or if null, updates the {@link action}
+     * 
+     * @param actionIn
+     */
+    @Nullable
+    public void updatePreAction(IEntityAction actionIn) {
```
maxanier: Is there any scenario where we would want to update a action that isn't `this.action`

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+        updateAction(null);
+    }
+
+    /**
+     * updates the given {@link IEntityAction} or if null, activates the {@link action}
+     * 
+     * @param actionIn
+     */
+    @Nullable
+    public void updateAction(IEntityAction actionIn) {
```
maxanier: Is there any scenario where we would want to update an action that isn't `this.action`?
Also void so no Nullable needed :)

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+        activateAction(null);
+    }
+
+    /**
+     * activates the given {@link IEntityAction} or if null, activates the {@link action}
+     * 
+     * @param actionIn
+     */
+    @Nullable
+    public void activateAction(IEntityAction actionIn) {
```
maxanier: Is there any scenario where we would want to activate an action that isn't `this.action`

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    /**
+     * returns a plausible action
+     * 
+     * elevates the chance of the actions, based on the entity and its target
+     * 
+     * @returns
+     */
+    private IEntityAction getIntelligentAction() {
```
maxanier: Maybe
`@NonNull` and `chooseNewAction`

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+        return action;
+    }
+
+    public boolean isPlayerTarget() {
+        return isPlayerTarget;
+    }
+
+    public void readFromNBT(NBTTagCompound nbt) {
+        if (nbt.hasKey("activeAction")) {
+            deactivateAction(VampirismAPI.entityActionManager().getRegistry().getValue(new ResourceLocation("vampirism", nbt.getString("activeAction"))));
```
maxanier: "Vampirism" probably should not be hardcoded here, since in theory other mods could register their own actions.
The modid should be part of the saved nbt string

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -0,0 +1,291 @@
...
+            isPlayerTarget = true;
+        }
+    }
+
+    public void writeToNBT(NBTTagCompound nbt) {
+        if (isPlayerTarget() && getAction() != null) {
+            nbt.setString("activeAction", action.getRegistryName().getResourcePath());
```
maxanier: See above

src/main/java/de/teamlapen/vampirism/config/BalanceEntityActions.java
```diff
@@ -0,0 +1,100 @@
...
+
+import de.teamlapen.lib.lib.config.BalanceValues;
+import de.teamlapen.lib.lib.config.DefaultDouble;
+import de.teamlapen.lib.lib.config.DefaultInt;
+import de.teamlapen.vampirism.VampirismMod;
+import java.io.File;
+
+public class BalanceEntityActions extends BalanceValues {
+
+    @DefaultDouble(value = 0.3, name = "Disruption health amount", comment = "health points until action will be disrupted of max health")
```
maxanier: "Health amount which an entity has to loose to disrupt an action. Relative value to max health."?

Also maybe add min and max boundry.

maxanier (changes requested):
Nice. Mainly cosmetic suggestions now.

However, currently it does not seem possible for other mods to use the ActionHandler without extending EntityVampirism (probably none ever will, but who knows, maybe it will be useful for the integrations mod). EntityActionHandler requires a EntityVampirism because one or two of it's methods are used. It would be good if these methods would be moved to a separate API interface, so the action Handler would use the Generic `T extends EntityCreature & IEntityActionUser` where IEntityActionUser extends `IFacionEntity` and IAdjustableLevel and defines the required methods.

maxanier:
It would be awesome if the action to activate selection would not be hardcoded in the action handler, but somehow let the actions discuss which one should be selected. (E.g. by letting them return a weight depending on the entity state and then choose a random one while considering the weight).
However, I think that does not have to be implemented now. We could implement this later or in a followup PR

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
Cheaterpaul: if we have an entity that uses multiple actions at the same time. not yet.
but could be solved otherwise, if we need it.

src/main/java/de/teamlapen/vampirism/config/BalanceEntityActions.java
Cheaterpaul: exactly

maxanier (approved):

Issue #377, opened 2019-01-14
Hunter Skill/Action: Vampire Awareness


maxanier:
Suggestion by _KingOfCandy_
Lasting skill with long duration, but long cooldown (or no cooldown) not sure, that makes the screen border turn red/violett if there is a IVampire creature or vampire player within a certain block radius of the player.

Alternatvevly the border could light up for a second, every time a vampire npc detects the hunter and sets it's attack target

PR #394, opened 2019-03-10
new hunter action

Cheaterpaul:
action shows screen border when a vampire is in a 12 block radius sphere
new icon and color changeable
disguise overlay now disappear after action ended
#377


src/main/resources/assets/vampirism/lang/en_US.lang
```diff
@@ -112,6 +113,7 @@ action.vampirism.vampire.halfinvulnerable_skill=Damage Limiter
 
 #HUNTER ACTIONS
 action.vampirism.hunter.disguise=Disguise
+action.vampirism.hunter.awareness=Awareness
```
maxanier, line 116: Maybe vampire awareness

src/main/resources/assets/vampirism/lang/en_US.lang
```diff
@@ -84,6 +84,7 @@ text.vampirism.skill.holy_water_enhanced=Enhanced Holy Water
 text.vampirism.skill.holy_water_enhanced.desc=Create enhanced Holy Water instead of normal one.
 text.vampirism.skill.garlic_beacon_improved=Improved diffuser
 text.vampirism.skill.garlic_beacon_improved.desc=Learn how to craft a garlic diffuser with a higher range
+text.vampirism.skill.awareness=Notify if vampire Players or Entities are nearby
```
maxanier, line 87: Maybe  "Sense nearby vampire players or entities"

src/main/java/de/teamlapen/vampirism/player/hunter/actions/AwarenessHunterAction.java
```diff
@@ -0,0 +1,97 @@
...
+        if (nearbyVampire(player)) {
+            ((HunterPlayer) player).getSpecialAttributes().nearbyVampire();
+        } else {
+            ((HunterPlayer) player).getSpecialAttributes().resetVampireNearby();
+        }
+        return false;
+    }
+
+    private boolean nearbyVampire(IHunterPlayer player) {
+        List<EntityLivingBase> entities = player.getRepresentingEntity().getEntityWorld().getEntitiesWithinAABB(EntityLivingBase.class, new AxisAlignedBB(player.getRepresentingEntity().posX - Balance.hpa.AWARENESS_RADIUS, player.getRepresentingEntity().posY
```
maxanier, line 87: If it is easily possible to avoid creating a new BB every time that would be good, if it isn't feasible, that's fine too.

I think there is a hardcoded 12 left here

src/main/java/de/teamlapen/vampirism/player/hunter/actions/AwarenessHunterAction.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public void onReActivated(IHunterPlayer player) {
+        onUpdate(player);
+    }
+
+    @Override
+    public boolean onUpdate(IHunterPlayer player) {
+        if (nearbyVampire(player)) {
```
maxanier, line 78: Probably better to only update this only every 10 or 20 ticks to reduce load. (player.getRepresentingEntity.ticksExisted%20==0)

src/main/java/de/teamlapen/vampirism/player/hunter/HunterPlayerSpecialAttribute.java
```diff
@@ -32,4 +35,26 @@ public void resetDisguise() {
         disguiseTicks = 0;
         disguised = false;
     }
+
+    public void nearbyVampire() {
+        if (vampireNearby) {
+            vampireNearbyTicks++;
```
maxanier, line 41: Although quite unlikely, the int might overflow. If you cap it to 20 here, you don't have to do it in #getVampireNearbyProgress

src/main/java/de/teamlapen/vampirism/player/hunter/HunterPlayerSpecialAttribute.java
```diff
@@ -11,6 +11,9 @@
     private boolean disguised = false;
     private int disguiseTicks = 0;
 
+    private boolean vampireNearby = false;
```
maxanier, line 14: This can probably avoided by using vampireNearbyTicks == 0 instead of this.

maxanier (changes requested):
Feels awesome to sense those evil vampires :smile: 

I think the radius could be bigger though, maybe twice? Also the color seems almost pink, so maybe it could get a little bit more reddish.

Off topic. Can you branch future PR from the latest master or replace the `import jline.Nullable`  in `SimpleSpawnerLogic` with `javax.Nullable`. Somehow that package has been removed from my setup. Guess it has been removed from Forge

Cheaterpaul:
closed(fail) new on is coming up

Issue #396, opened 2019-03-12
Server Crash: Choose new entity action


maxanier:
Time: 3/12/19 1:49 AM
Description: Ticking entity

java.lang.IllegalArgumentException
        at net.minecraft.util.WeightedRandom.func_76273_a(SourceFile:18)
        at net.minecraft.util.WeightedRandom.func_76271_a(SourceFile:37)
        at de.teamlapen.vampirism.entity.action.EntityActionHandler.chooseNewAction(EntityActionHandler.java:148)
        at de.teamlapen.vampirism.entity.action.EntityActionHandler.updateHandler(EntityActionHandler.java:129)
        at de.teamlapen.vampirism.entity.action.EntityActionHandler.handle(EntityActionHandler.java:155)
        at de.teamlapen.vampirism.entity.EntityVampirism.func_70636_d(EntityVampirism.java:160)
        at de.teamlapen.vampirism.entity.vampire.EntityVampireBase.func_70636_d(EntityVampireBase.java:190)
        at de.teamlapen.vampirism.entity.vampire.EntityBasicVampire.func_70636_d(EntityBasicVampire.java:134)
        at net.minecraft.entity.EntityLivingBase.func_70071_h_(EntityLivingBase.java:2179)
        at net.minecraft.entity.EntityLiving.func_70071_h_(EntityLiving.java:295)
        at de.teamlapen.vampirism.entity.EntityVampirism.func_70071_h_(EntityVampirism.java:166)
        at de.teamlapen.vampirism.entity.vampire.EntityBasicVampire.func_70071_h_(EntityBasicVampire.java:153)
        at net.minecraft.world.World.func_72866_a(World.java:1993)
        at net.minecraft.world.WorldServer.func_72866_a(WorldServer.java:832)
        at net.minecraft.world.World.func_72870_g(World.java:1955)
        at net.minecraft.world.World.func_72939_s(World.java:1759)
        at net.minecraft.world.WorldServer.func_72939_s(WorldServer.java:613)
        at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:767)
        at net.minecraft.server.dedicated.DedicatedServer.func_71190_q(DedicatedServer.java:397)
        at net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:668)
        at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:526)
        at java.lang.Thread.run(Thread.java:748)

[...]

PR #397, opened 2019-03-13
Server Crash: Choose new entity action Bugfix

Cheaterpaul:
now skips actions with weight 0
now writes EntityClassType to nbt
#396

### DIFF
--- a/src/api/java/de/teamlapen/vampirism/api/entity/EntityClassType.java
+++ b/src/api/java/de/teamlapen/vampirism/api/entity/EntityClassType.java
@@ -1,6 +1,8 @@
 package de.teamlapen.vampirism.api.entity;
 
+import com.google.common.collect.Maps;
 import net.minecraft.entity.ai.attributes.AttributeModifier;
+import java.util.Map;
 import java.util.Random;
 
 public enum EntityClassType {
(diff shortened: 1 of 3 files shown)


src/api/java/de/teamlapen/vampirism/api/entity/EntityClassType.java
```diff
@@ -35,4 +40,19 @@ public AttributeModifier getDamageModifier() {
     public AttributeModifier getSpeedModifier() {
         return speedModifier;
     }
+
+    public static int getID(EntityClassType entityclasstypes) {
+        return ENTITYCLASSTYPES.get(entityclasstypes);
+    }
+
+    public static EntityClassType getEntityClassType(int id) {
+        return ID.get(id);
```
maxanier: To be sure this should probably be checked against map.size()  and return a default if to large

src/main/java/de/teamlapen/vampirism/entity/action/EntityActionHandler.java
```diff
@@ -139,13 +142,20 @@ private void updateHandler() {
...
         List<EntityActionEntry> entry = Lists.newArrayList();
+        int weightsum = 0;
         for (IEntityAction e : availableActions) {
-            entry.add(new EntityActionEntry(((DefaultEntityAction<T>) e).getWeight(entity), e));
+            int weight = ((DefaultEntityAction<T>) e).getWeight(entity);
+            if (weight > 0) {
+                entry.add(new EntityActionEntry(weight, e));
+            }
+        }
+        if (weightsum > 0) {
```
maxanier, line 156: `weightsum` is never increased :)

maxanier (changes requested):

src/api/java/de/teamlapen/vampirism/api/entity/EntityClassType.java
Cheaterpaul: dont set an classtype if id is corrupt(entityclasstype is set by contructor)

PR #398, opened 2019-03-13
Vampire awareness new

Cheaterpaul:
Superseeds #394 
(branch pull/push conflicts)
1. Commit the same as last PR
2. changes

modifying the AxisAlignedBB ist possible with moving it every tick(using prevPosX&Y&Z) or usage (must save last position of usage) if its still needed
#377


maxanier:
Excellent.
Last thing to discuss is the placement of the skill in the skill tree.
Since the middle branch is quite unpopular right now, I would suggest moving it to the end of the left one or as an alternative to one at the right side

PR #401, opened 2019-03-21
huntertrainer is now no iHunter

Cheaterpaul:
look for safety


maxanier:
One issue with this is that vampires will no longer avoid (or attack) the trainer while it will still attack them.

Issue #404, opened 2019-04-01
Blood vision upgrade: See garlic


maxanier:
As an alternative to "Vista" there could be a skill that upgrades blood vision to also outline garlic infused entities (hunter/villagers/...) in blue/white

TheDoomSlayer:
hello. I made new models for the mod.

PR #405, opened 2019-04-02
blood vision garlic infused entities

Cheaterpaul:
#404 
vampire skill action blood vision now outline hunter & entities with poisonous blood in blue


maxanier:
Cool, that was quick.
Does this already cover hunter player?

Issue #415, opened 2019-05-07
[Feature] obsidian armor change


Nashor-Senpai:
Could you perhaps change the poison in obsidian armor  for vampires from poison II to garlic poison or any other kind of damaging debuff? there are several mods like grimoire of gaia or baubles that allows you to farm poison immune trinkets, that would allow vampires to wear obsidian armor and straight giving them a fair advantage.
Also, it would be great if there were a way to craft special armor as its suposed to give hunters an advantage against vampires but if you cant even find them and vampires can equip them with poison immunity it doesnt feel that rewarding.

Cheaterpaul:
#422

PR #424, opened 2019-05-18
Modded Poison & some stuff

Cheaterpaul:
adding modded poison potion
changing to modded poison for internal damage
changing name for vampire cloak
adding item description
#415


src/main/java/de/teamlapen/vampirism/client/gui/VampirismHUDOverlay.java
```diff
@@ -55,6 +57,7 @@
 
     private final Minecraft mc;
     private final ResourceLocation icons = new ResourceLocation(REFERENCE.MODID + ":textures/gui/icons.png");
+    private final ResourceLocation icons_poison = new ResourceLocation(REFERENCE.MODID, "textures/gui/icons_poison.png");
```
maxanier, line 60: If you want to use the original item, please bind the MC texture or add the icon to our custom file.

src/main/resources/assets/vampirism/lang/en_US.lang
```diff
@@ -356,6 +352,7 @@ effect.vampirism.saturation=Saturation
 effect.vampirism.sanguinare=Sanguinare Vampiris
 effect.vampirism.fire_protection=Fire Protection
 effect.vampirism.garlic=Garlic
+effect.vampirism.poison=Poison
```
maxanier: Maybe change the name to make it clear that this is not the Vanilla poison.
Maybe "Vampirism poison" or "Toxicant"

maxanier (changes requested):

Issue #427, opened 2019-05-19
Decrease yellow sunborder screen effect when using sunscreen


maxanier:
If taking sundamage is prevented the border does not need to be that annoying

Cheaterpaul:
possibly remove the sun border in creative mode?

maxanier:
No, it should stay as is (reduced version).
Useful to debug things, build vampire areas in creative and more.
Users can always completely disable the sunborder if they want

PR #429, opened 2019-05-20
smaller sun screen percentage when sunscreen action active

Cheaterpaul:
#427


maxanier:
Since lower levels of the potion only reduce the damage dealt, I adjusted your change to only consider amplifier >=4

Issue #428, opened 2019-05-20
Turned Vampire Villager Trades[Bug]


princessbinas:
## Versions
- Minecraft: 1.12.2                
- Forge: 14.23.5.2838                
- Vampirism: 1.5.0-beta.1                                 

## Issue Description

When trading with villagers to obtain blood bottles and hearts, the villagers do not give them despite showing the trades. 

## Reproduce Steps

- Obtain emeralds
- Find a villager (preferably a vanilla villagers such as a farmer)
- Have the villager turned into a vampire
- Scroll to a trade that has either "X emeralds for X blood bottles" or "X emeralds for X hearts." It does not matter as it happens with both trades.
- Attempt to trade for the blood bottles and/or hearts
- If the villager was a farmer, you will get bread most likely. Another vampire villager type (i.e. Blacksmith) might give you something different if you plugged in the right amount of emeralds. If the villager does not have a "X emeralds for X" trade, the villager will reject your offer.
- Exit out of the villager trade menu
- The trade disappears all together on the villager and never reappears no matter how many times you trade with the villager.

## Additional Information

I have tested this bug several times in different scenarios. This bug is present in both single player and on the official vampire server. No other mods seem to play a role in this bug and the forge version does not seem to matter. This bug is also consistent in how it happens. It always happens.

Cheaterpaul:
my research until now:
these bugs appear only in the official vampirism server modpack, even if you delete all mods except Vampirism
edit:
the forge version is the problem

Cheaterpaul:
i opened another issue for this problem (#434) and closing this, because first this is a issue of the modpack (vampirism has the issue with the forge version 14.23.5.2838/2836, while vampirism's forge 14.23.5.2768 works as intended.

I have already reported the problem on the server discord

lunofe:
Modpack has been updated, it should work now.

maxanier:
@Cheaterpaul But the issue still exists with the latest Forge version? So we need to find out what exactly is causing this and if it has to be fixed on Vampirism's side or Forge's side. Downgrading Forge is just a workaround.
Or did I misunderstand something?

Cheaterpaul:
you are right, i opened #434 , but we can continue here

princessbinas:
I know that the previous official release of Vampirism (1.4.9) did not have this issue, even with the forge version documented in my bug report and on wards. I am not sure how much this information will help.

[...]

PR #435, opened 2019-05-27
villager trades

Cheaterpaul:
adding recipes directly to buying list #428


maxanier:
Closed  by 869ac46ca95076dec21105c046517cc7b14d18a0
We can add the items on conversion, no need to wait until later. Makes it a little bit clearer

PR #436, opened 2019-05-27
added a dedicated class for converted horses

### LINKED ISSUE #426: [Bug] Make vampire horses rideable
Probably require a dedicated converted horse class

Cheaterpaul:
Closes #426

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedHorse.java
+++ b/src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedHorse.java
@@ -1,17 +1,20 @@
 package de.teamlapen.vampirism.entity.converted;
 
+import de.teamlapen.lib.lib.util.UtilLib;
 import de.teamlapen.vampirism.api.EnumStrength;
 import de.teamlapen.vampirism.api.VReference;
 import de.teamlapen.vampirism.api.VampirismAPI;
 import de.teamlapen.vampirism.api.entity.convertible.IConvertedCreature;
 import de.teamlapen.vampirism.api.entity.convertible.IConvertingHandler;
+import de.teamlapen.vampirism.api.items.IVampireFinisher;
+import de.teamlapen.vampirism.config.Balance;
 import de.teamlapen.vampirism.core.ModPotions;
 import de.teamlapen.vampirism.entity.DamageHandler;
+import de.teamlapen.vampirism.entity.EntityCrossbowArrow;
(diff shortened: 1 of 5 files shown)


src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedHorse.java
```diff
@@ -0,0 +1,105 @@
...
+    public EnumStrength isGettingGarlicDamage(boolean forceRefresh) {
+        if (forceRefresh) {
+            garlicCache = Helper.getGarlicStrength(this);
+        }
+        return garlicCache;
+    }
+
+    @Override
+    public boolean isGettingSundamage(boolean forceRefresh) {
+        // TODO Auto-generated method stub
```
maxanier: Guess vampire horses should also get sundamage

src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedHorse.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public boolean useBlood(int amt, boolean allowPartial) {
+        this.addPotionEffect(new PotionEffect(MobEffects.WEAKNESS, amt * 20));
+        bloodTimer = 0;
+        return true;
+    }
+
+    @Override
+    public boolean wantsBlood() {
+        return bloodTimer > 0;
```
maxanier: But blood timer is never increased.

Also, since they will never suck blood, we probably do not need this at all

maxanier (changes requested):

maxanier:
Sundamage has to be actively checked and potentially dealt in the entities update method, otherwise it does not have any affect. Vampire horses should behave similar to normal converted creatures

maxanier:
Why are the horses only getting weakness while other converted animals die?
Since they might be used by players, this might make it a little easier to use them, but still it is inconsistent. Players should take care of their vampire horses. Only compromise would be giving them only very little sundamage, so players have sufficient time to bring them into safety.

maxanier:
Did you even test this?

a) The vampire horses don't have a vampire eye overlay anymore (No renderer is registered)
b) The entity name language key/translation is missing

It is awesome that you are contributing to the mod and doing so many PRs, but pls make sure they actually work, otherwise it would be faster if I implemented them myself (as testing is usually the part that takes the longest).

maxanier:
Entity name was still wrong
Also converted horses could be bitten after reloading the world

PR #441, opened 2019-06-07
abort capture before remove the tile

Cheaterpaul:
(Intention: replace aggressive villagers with normal)


src/main/java/de/teamlapen/vampirism/tileentity/TileTotem.java
```diff
@@ -392,7 +392,10 @@ public void onTileRemoved() {
         if (v != null) {
             v.removeTotemAndReset(this.pos);
         }
-        updateBossinfoPlayers(null);
+        if (capturingFaction != null)
```
maxanier, line 395: Pls only skip the curly brackets for one liners. With if/else it might be missed at some point

PR #451, opened 2019-06-18
TODOs

Cheaterpaul:
TODOs for all errors except modcompat & tileentity packages

fixed some & created spawneggs for entities

for TODOswith @maxanier/? I have no idea, or you should do it or i need another look

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedCreature.java
+++ b/src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedCreature.java
@@ -1,7 +1,6 @@
 package de.teamlapen.vampirism.entity.converted;
 
 import de.teamlapen.lib.lib.network.ISyncable;
-import de.teamlapen.lib.lib.util.UtilLib;
 import de.teamlapen.vampirism.api.VReference;
 import de.teamlapen.vampirism.api.VampirismAPI;
 import de.teamlapen.vampirism.api.entity.convertible.IConvertedCreature;
@@ -18,6 +17,7 @@ import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.util.text.ITextComponent;
 import net.minecraft.util.text.TextComponentString;
+import net.minecraft.util.text.TextComponentTranslation;
 import net.minecraft.world.World;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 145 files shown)


src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedCreature.java
```diff
@@ -258,12 +259,12 @@ private void writeOldEntityToNBT(NBTTagCompound nbt) {
         if (!nil()) {
             try {
                 NBTTagCompound entity = new NBTTagCompound();
-                entityCreature.isDead = false;
+                entityCreature.dead = false;//TODO is protected
```
maxanier: Careful. This is a completely different variable than the old one. (Different class, different effect)
The original Entity#isDead is now inverted and renamed to Entity#isAlive (or respectively removed).
The only intention is to trick the check in Entity#writeUnlessPassenger

src/main/java/de/teamlapen/vampirism/entity/converted/EntityConvertedCreature.java
```diff
@@ -42,7 +43,7 @@ public EntityConvertedCreature(EntityType type, World world) {
 
     @Override
     public ITextComponent getName() {
-        return UtilLib.translate("entity.vampirism.vampire.name") + " " + (nil() ? super.getName() : entityCreature.getName());
+        return new TextComponentString(UtilLib.translate("entity.vampirism.vampire.name") + " " + (nil() ? super.getName() : entityCreature.getName()));
```
maxanier: Why don't we use TextComponentTranslation

src/main/java/de/teamlapen/vampirism/entity/minions/SaveableMinionHandler.java
```diff
@@ -90,13 +90,13 @@ public NBTTagList getMinionsToSave() {
         NBTTagList list = new NBTTagList();
         for (IMinion m : minions) {
             Entity e = MinionHelper.entity(m);
-            boolean dead = e.isDead;
-            e.isDead = false;
+            boolean dead = e.dead;//TODO protected attribute
```
maxanier: same here

src/main/java/de/teamlapen/vampirism/items/ItemBloodPotion.java
```diff
@@ -76,7 +76,7 @@ public ItemStack onItemUseFinish(ItemStack stack, World worldIn, EntityLivingBas
             BloodPotions.applyEffects(stack, entityLiving);
         }
 
-        if (entityplayer == null || !entityplayer.isCreative()) {
+        if (entityplayer == null || !entityplayer.abilities.isCreativeMode) {
```
maxanier, line 79: Oh, I didn't know that one still existed, but what is the advantage of that one vs isCreative()

maxanier (commented):
I am not really see the necessity to add TODO comments to unchanged compile errors, since the compiler already tells me where they are? But ok

maxanier:
Awesome, thank you

PR #452, opened 2019-06-19
entityactions (request)

Cheaterpaul:
I personally think it's not a nice coding
edit: another idea is a subclass of EntityVampirism which the entityactionuser extends insead of EntityVampirism


src/main/java/de/teamlapen/vampirism/entity/vampire/EntityBasicVampire.java
```diff
@@ -366,4 +386,31 @@ protected void updateEntityAttributes() {
...
+    }
+
+    @Override
+    public EntityActionTier getEntityTier() {
+        return entitytier;
+    }
+
+    @Override
+    public List<IEntityAction> getAvailableActions() {
+        return VampirismAPI.entityActionManager().getAllEntityActionsByTierAndClassType(((IFactionEntity) this).getFaction(), entitytier, entityclass);
```
maxanier, line 406: Could maybe be moved to the API as a default implementation.
Even though the API should contain as few actual code as possible, I think this could be ok

maxanier:
Additional subclass would be nicer, but since we already have separate subclasses for vampire and hunter entities, that probably won't work well.
Yes, it is some duplicate code. But the vanilla MC code has so much duplicates, we fit right in

maxanier:
Pls resolve conflicts

maxanier:
Thank you

PR #464, opened 2019-07-04
1.13 stuff

### LINKED ISSUE #463: [1.13] Crashes
-SkillEvent.CreatedNode is in catched

Cheaterpaul:
advancement fix
#463 fix (eventually only temporary, need lots of testing (do when testing everything))
(first crash)
(IllegalStateException when getting extended creature from a vampire)


Cheaterpaul:
https://github.com/TeamLapen/Vampirism/blob/eeb6dfd1d6faa63c00b2fe3c0a63eaedb1f42142/src/lib/java/de/teamlapen/lib/network/UpdateEntityPacket.java#L63
return pkt?

src/main/java/de/teamlapen/vampirism/entity/ai/VampireAIBiteNearbyEntity.java
```diff
@@ -45,7 +45,7 @@ public boolean shouldContinueExecuting() {
     @Override
     public boolean shouldExecute() {
         if (vampire.wantsBlood()) {
-            List<EntityCreature> list = vampireEntity.getEntityWorld().getEntitiesWithinAABB(EntityCreature.class, getBiteBoundingBox());
+            List<EntityCreature> list = vampireEntity.getEntityWorld().getEntitiesWithinAABB(EntityCreature.class, getBiteBoundingBox(), (entityCreature -> !(entityCreature instanceof IVampireMob)));
```
maxanier, line 48: This prevents the crash, but does not resolve the root issue.
Even IVampireMobs should have a ExtendedCreature capability shouldn't they? (Even though it is not used)

Cheaterpaul: I have tested it with multiple tries and creature entities and it has never throws an error. So this works until someone finds the basic problem or gets the IllegalStateException with creatures other than vampires

Cheaterpaul: The exception will not be thrown the first time get() is called in ExtendedCreature, but it will be triggered randomly after some successful calls, also only crashes if the vampire checks itself

maxanier:
> https://github.com/TeamLapen/Vampirism/blob/eeb6dfd1d6faa63c00b2fe3c0a63eaedb1f42142/src/lib/java/de/teamlapen/lib/network/UpdateEntityPacket.java#L63
> 
> return pkt?

Absolutely :D Fixed, now UpdateEntityPaket is working

PR #478, opened 2019-07-18
recipes and loottable, some changes

Cheaterpaul:
now able join a world
- except the configs


src/main/java/de/teamlapen/vampirism/command/test/SkillCommand.java
```diff
@@ -53,7 +54,7 @@ private static int skill(CommandSource commandSource, ServerPlayerEntity asPlaye
             (factionPlayer.getSkillHandler()).resetSkills();
             return 0;
         }
-        ISkill skill = ModRegistries.SKILLS.getValue(new ResourceLocation(type));
+        ISkill skill = ModRegistries.SKILLS.getValue(new ResourceLocation(REFERENCE.MODID, type));
```
maxanier, line 57: Compatibility for addons could be done like vanilla does.
If no resource domain is given, vampirism is used otherwise the given one is used

src/main/java/de/teamlapen/vampirism/entity/VampirismEntity.java
```diff
@@ -316,4 +319,30 @@ protected void updateAITasks() {
             onRandomTick();
         }
     }
+
+    public static boolean spawnPredicateVampire(EntityType<? extends VampirismEntity> entityType, IWorld world, SpawnReason spawnReason, BlockPos blockPos, Random random) {
```
maxanier, line 323: Why are these in VampirismEntity and not in the according subclass?

PR #501, opened 2019-08-18
modifiable SelectActionMenu

Cheaterpaul:
open editable SelectActionMenu from SkillScreen


Cheaterpaul:
this is the easiest way to do it, i could add a new gui instead of SelectActionScreen for modification of the order.

src/main/java/de/teamlapen/vampirism/VampirismMod.java
```diff
@@ -267,6 +268,7 @@ private void setupClient(FMLClientSetupEvent event) {
         registryManager.onInitStep(IInitListener.Step.CLIENT_SETUP, event);
         proxy.onInitStep(IInitListener.Step.CLIENT_SETUP, event);
         modCompatLoader.onInitStep(IInitListener.Step.CLIENT_SETUP, event);
+        SelectActionScreen.loadActionOrder();
```
maxanier: Klappt das auch aufm dedicated server, weil SelectActionScreen kann nur auf dem client geladen werden, aber setupClient wird aufm server geladen (nur nicht ausgeführt) glaube ich.

Falls nicht -> client proxy

maxanier:
This is amazing :) Easy to use and looks cool.

Only thing left todo is add some kind of tutorial on how to use it. Maybe a test message that is sent when the player opens the edit screen? Or something cooler

maxanier:
Even works quite well with multiple worlds

PR #503, opened 2019-08-21
bloodvalues in datapack

Cheaterpaul:
Now loades fluid conversion rates, food to blood values and entity blood vales from datapack.
Fluids needs to be completed.


src/api/java/de/teamlapen/vampirism/api/entity/IVampirismEntityRegistry.java
```diff
@@ -101,15 +101,7 @@
     @Nullable
     BiteableEntry getEntry(CreatureEntity creature);
 
+    void processBloodValues(int multiplier);
```
maxanier: Nobody except Vampirism should ever use this, so no need to have this in the API right? Just cast the registry if necessary

src/main/resources/data/vampirism/vampirism_blood_values/fluids/minecraft.txt
```diff
@@ -0,0 +1,2 @@
+#Divider
+minecraft:divider
```
maxanier: Is this supposed to not have a value?

maxanier (commented):
If I understand correctly the proper way for other mods to add values for their own mod is to have a datapack file. If that's correct we probably don't need the 
    @ThreadSafeAPI
    void addBloodValue(ResourceLocation entityId, int value);

    /**
     * Add a set of entity id -> blood value pairs
     */
    @ThreadSafeAPI
    void addBloodValues(Map<ResourceLocation, Integer> values);

in the API anymore. And since no values are ever added individually 
    void processBloodValues(int multiplier);
    void prepareBloodValues();
    addBloodValues()

could be combined into a single function.
Maybe the same applies to the fluid and item registry

PR #511, opened 2019-09-15
1.14 fluids & fixes & some stuff

Cheaterpaul:
fluids should work for 99% (there could exist some != null checks that can be removed/ replaced by .isEmpty() ), water in blood container still rendered without color, extracting fluid out of container still doesnt work(waiting for forge PR)

other changes listed in commits

have fun 😃


src/lib/java/de/teamlapen/lib/lib/util/NotDrainableTank.java
```diff
@@ -0,0 +1,37 @@
+package de.teamlapen.lib.lib.util;
+
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.fluids.capability.templates.FluidTank;
+
+import javax.annotation.Nonnull;
+import java.util.function.Predicate;
+
+public abstract class NotDrainableTank extends FluidTank {
```
maxanier: Name is quite misleading.
Could also be combined with the FluidTankWithListener

src/main/java/de/teamlapen/vampirism/blocks/CastleBricksBlock.java
```diff
@@ -54,35 +54,14 @@ public EnumVariant getVariant() {
...
-
+        DARK_BRICK(),
+        PURPLE_BRICK(),
+        DARK_BRICK_BLOODY(),
+        NORMAL_BRICK(),
+        DARK_STONE();
 
         public String getName() {
-            return this.name;
+            return name().toLowerCase();
```
maxanier, line 64: Not sure if this is better. Since it requires a new string object every time the method is called. But ok

src/main/java/de/teamlapen/vampirism/blocks/BloodContainerBlock.java
```diff
@@ -87,34 +95,31 @@ public boolean isNormalCube(BlockState state, IBlockReader worldIn, BlockPos pos
...
-                        playerIn.setHeldItem(hand, result.getResult());
-                    }
-                    //fillContainerFromTank(worldIn, pos, playerIn, stack, bloodContainer);
-                } else {
-                    FluidActionResult result = FluidUtil.tryEmptyContainer(stack, source, Integer.MAX_VALUE, playerIn, true);
-                    //drainContainerIntoTank(worldIn, pos, playerIn, stack, bloodContainer);
-                    if (result.isSuccess()) {
-                        playerIn.setHeldItem(hand, result.getResult());
-                    }
+        FluidUtil.interactWithFluidHandler(playerIn, hand, worldIn, pos, hit.getFace());
```
maxanier: Isn't this line doing something similar to the remaining part?

src/main/java/de/teamlapen/vampirism/blocks/BloodContainerBlock.java
```diff
@@ -135,6 +140,59 @@ public void onBlockPlacedBy(World worldIn, BlockPos pos, BlockState state, Livin
         }
     }
 
+    public static FluidActionResult tryEmptyContainer(@Nonnull ItemStack container, IFluidHandler fluidDestination, int maxAmount, @Nullable PlayerEntity player, boolean doDrain) {
```
maxanier: How are these methods different from the FluidUtil ones?

src/main/java/de/teamlapen/vampirism/config/VampirismConfig.java
```diff
@@ -130,7 +130,7 @@ public static void onReload(final ModConfig.ConfigReloading configEvent) {
             sunscreenBeaconDistance = builder.comment("Block radius the sunscreen beacon affects").defineInRange("sunscreenBeaconDistance", 32, 1, 40000);
             sunscreenBeaconMineable = builder.comment("Whether the suncreen beacon can be mined in survival").define("sunscreenBeaconMineable", false);
             autoCalculateEntityBlood = builder.comment("Calculate the blood level for unknown creatures based on their size").define("autoCalculateEntityBlood", true);
-            autoConvertGlassBottles = builder.comment("Whether glass bottles should be automatically be converted to blood bottles when needed").define("autoConvertGlassBottles", true);
+            autoConvertGlassBottles = builder.comment("Whether glass bottles should be automatically be converted to blood bottles when needed").define("autoConvertGlassBottles", false);
```
maxanier: Any specific reason?

src/main/java/de/teamlapen/vampirism/tileentity/CoffinTileEntity.java
```diff
@@ -14,25 +14,37 @@
...
 import javax.annotation.Nullable;
 
 /**
  * TileEntity for coffins. Handles coffin lid position and color
  */
 public class CoffinTileEntity extends TileEntity implements ITickableTileEntity {
     public int lidPos;
-    public int color = 15;
+    public DyeColor color = DyeColor.BLACK;
+    public final boolean item;
```
maxanier: Not a very meaningful attribute name

src/main/java/de/teamlapen/vampirism/core/ModEntities.java
```diff
@@ -90,79 +91,85 @@ static void registerCustomExtendedCreatures() {
...
         registry.addConvertible(EntityType.VILLAGER, VillagerEntity.class, null, new ConvertedVillagerEntity.ConvertingHandler());
-        registry.addConvertible(EntityType.LLAMA, LlamaEntity.class, String.format(base, "llama"));
-        registry.addConvertible(EntityType.PANDA, PandaEntity.class, String.format(base, "panda"));
     }
 
     static void registerEntities(IForgeRegistry<EntityType<?>> registry) {
         //simply register EntityType
-        registry.register(blinding_bat);
-        registry.register(ghost);
+        registry.register(advanced_hunter);
```
maxanier, line 111: A little bit inconsistent with some elements being initialized in static and others here. But ok

src/main/java/de/teamlapen/vampirism/core/ModEventHandler.java
```diff
@@ -125,6 +133,12 @@ public void onPlayerLoggedIn(PlayerEvent.PlayerLoggedInEvent event) {
         }
         VampirismMod.dispatcher.sendTo(new SkillTreePacket(VampirismMod.proxy.getSkillTree(false).getCopy()), (ServerPlayerEntity) event.getPlayer());
 
+        Map<ResourceLocation, Pair<Map<ResourceLocation, Integer>, Integer>> bloodValues = Maps.newConcurrentMap();
+        bloodValues.put(new ResourceLocation(REFERENCE.MODID, "entities"), new Pair<>(((VampirismEntityRegistry) VampirismAPI.entityRegistry()).getBloodValues(), ((VampirismEntityRegistry) VampirismAPI.entityRegistry()).getBloodMultiplier()));
```
maxanier: Do we need a resource location with mod id here? Blood values added by other mods should be synchronized by Vampirism anyway so there should be no need for addons to use this or the mod id every being different

src/main/java/de/teamlapen/vampirism/items/BloodBottleFluidHandler.java
```diff
@@ -104,8 +101,24 @@ public ItemStack getContainer() {
...
+
+    @Nonnull
+    @Override
+    public FluidStack getFluidInTank(int tank) {
+        return new FluidStack(ModFluids.blood, getBlood(container));
+    }
+
+    @Override
+    public int getTankCapacity(int tank) {
+        return 0;
```
maxanier: Is this correct?

src/main/java/de/teamlapen/vampirism/tileentity/AltarInspirationTileEntity.java
```diff
@@ -156,22 +150,23 @@ public void tick() {
...
+            super(capacity, fluidStack -> ModFluids.blood.isEquivalentTo(fluidStack.getFluid()));
+            setDrainable(false);
         }
 
-        public FluidStack doDrain(int maxDrain, boolean doDrain) {
-            this.setCanDrain(true);
-            FluidStack s = super.drain(maxDrain, doDrain);
-            this.setCanDrain(false);
+        @Override
+        public FluidStack drain(int maxDrain, FluidAction action) {
```
maxanier: If I understand this correctly, this makes the NotDrainableTank quite pointless :D If you override the drain method, then every call to drain this tank will make it drainable and therefore void the drainable check of the original method.

In the old code this was done in a new dedicated method doDrain which is only called by Vampirism

maxanier (changes requested):
Awesome. Getting closer to an alpha release :)

src/main/java/de/teamlapen/vampirism/core/ModEntities.java
Cheaterpaul: intention consistency:
all forge registry items should be created in the register() Method.
the objects that must be used before forge assiges them to the Objectholder are created in the static initializer ( e.g. entitytypes for spawn eggs)
-> all attribute objectholder are short (no object creation)

src/main/java/de/teamlapen/vampirism/blocks/BloodContainerBlock.java
Cheaterpaul: maybe read the whole docs to know, that in creative mode the item doesnt get replaced if emptied

src/main/java/de/teamlapen/vampirism/config/VampirismConfig.java
Cheaterpaul: tested something, but didnt reset this

src/main/java/de/teamlapen/vampirism/blocks/BloodContainerBlock.java
Cheaterpaul: removed

Cheaterpaul:
no need to wait for the ForgePR, FluidUtil#interactWithFluidHandler() doesnt need it

maxanier:
Perfect :+1:

PR #515, opened 2019-09-24
grinder now supports items from other mods

Cheaterpaul:
(only meat, cooked meat should not be supported)

- calculated blood values for items
- reorder bloodvalueloader


maxanier:
Cool :+1: 

From the code it seems like cooked meat can result in some blood?

Also there should be a upper limit, just in case some mod adds some kind of full food bar item

PR #595, opened 2020-02-15
1.14 village

### LINKED ISSUE #594: [Bug][1.14] Hunter Trainer -> Undercover Huntertrainer + Vampire
when force converting a village to a vampire village. the hunter trainer turns into an undercover hunter trainer and an additional vampire spawn at the same position

Cheaterpaul:
- adds hunter trainer building to village
- adds totem to village
- ensure single totem/hunter trainer per village
- include LootHandler.class in ModLootTables.class

notes:
- hunter trainer building could need a workover
- spawn the buildings with the `minecraft:structure_block` block, select load mode and use:
`vampirism:village/plains/houses/hunter_trainer`,
`vampirism:village/desert/houses/hunter_trainer`,
`vampirism:village/savanna/houses/hunter_trainer`,
`vampirism:village/taiga/houses/hunter_trainer`,
`vampirism:village/snowy/houses/hunter_trainer`,
`vampirism:village/totem`


src/main/java/de/teamlapen/vampirism/config/BalanceConfig.java
```diff
@@ -228,6 +231,9 @@
         viPhase1Duration = builder.comment("Duration of phase 1 of the capturing process in 2*seconds").defineInRange("phase1Duration", 80, 1, 1000);
         viNotifyDistanceSQ = builder.comment("Squared distance of village capture notification").defineInRange("notifyDistanceSQ", 40000, 0, 100000);
         viForceTargetTime = builder.comment("Time in 2*seconds in capture phase 2 after which the capture entities should find a target regardless of distance").defineInRange("forceTargetTime", 80, 1, 1000);
+        viHunterTrainerWeight = builder.comment("Weight of the Hunter Trainer Building inside the village").defineInRange("viHunterTrainerWeight", 400, 1, Integer.MAX_VALUE);
+        viTotemWeight = builder.comment("Weight of the Totem Building inside the Village").defineInRange("viTotemWeight", 20,1,Integer.MAX_VALUE);
+        viTotemPreSetPercentage = builder.comment("Percentage of totem which should be have a pre defined faction").defineInRange("viTotemPreSetPercentage", 0.90d,0d,1d);
```
maxanier: Not sure how this was before, but I guess something like 50-60% would be more reasonable

src/main/java/de/teamlapen/vampirism/entity/hunter/HunterTrainerEntity.java
```diff
@@ -75,6 +78,22 @@ public void setHome(AxisAlignedBB box) {
         this.setMoveTowardsRestriction(MOVE_TO_RESTRICT_PRIO, true);
     }
 
+    @Override
+    public void readAdditional(CompoundNBT nbt) {
+        super.readAdditional(nbt);
+        if(nbt.contains("createHome") && (this.shouldCreateHome = nbt.getBoolean("createHome"))){
```
maxanier, line 84: Does this need to happen on load, is it not possible to set the home on entity creation?

Should "createHome" stay true after setting the home pos?

src/main/resources/META-INF/coremods/batmode.js
```diff
@@ -8,6 +8,7 @@ function initializeCoreMod() {
                 'methodDesc': '(Lnet/minecraft/entity/Pose;)Lnet/minecraft/entity/EntitySize;'
             },
             'transformer': function (method) {
+                Java.type('net.minecraftforge.coremod.api.ASMAPI').log("INFO", "HI!!");
```
maxanier: Can be removed. I just forgot it

maxanier (commented):
Awesome.
The snowy hunter house looks nice and cosy :)

The totem currently always uses grass blocks regardless of the biome right?
Is there a fancy way to adjust this code wise or is it simpler to just have different nbts for the different biomes.

Yes StructureManager can be removed

src/main/java/de/teamlapen/vampirism/entity/hunter/HunterTrainerEntity.java
Cheaterpaul: 1. not possible, because the entity is loaded from NBT (as save file), which results in not create a new Entity

2. the entity is loaded and saved two times before joining ?! -> if not saved the home is not created the second time.

3. if the home is set, it is saved by VampirismEntity and on the next load the home will not be created new

maxanier:
:bowing_man: :ok_hand:

PR #602, opened 2020-02-23
1.15 datagenerator

Cheaterpaul:
adding loottable datagenerator
adding tag datagenerator
(includes the village updates from 1.14, only the last commit brings the generators)

note:
hunter tents drops a hunter tent item - should remove?
its not possible to get ultimate vampire swords (only from vampire dungeon), while hunter can get all ultimate equipment


maxanier:
Nice.

Hunter tent should drop item (the non spawner one).
Vampire swords will have to wait

Cheaterpaul:
is there a specific reason why the blockstates are still in the main project?

maxanier:
Not all of them have been converted to datagenerators, so they must remain.
Didn't really see a reason to migrate to data generators for stuff that still works.

PR #621, opened 2020-03-07
1.14 sleeping in tents

### LINKED ISSUE #98: Sleep in tents
Players should be able to sleep in hunter tents

Cheaterpaul:
- adds the ability to sleep inside of tents as hunter
- adjusts the shape of the tent to include the backside
- close #98


src/main/java/de/teamlapen/vampirism/blocks/TentBlock.java
```diff
@@ -1,102 +1,293 @@
...
+            case 0:
+            case 3:
+                return state.get(HORIZONTAL_FACING).getOpposite();
+            default:
+                return state.get(HORIZONTAL_FACING);
+        }
+    }
+
+    @Override
+    public void onReplaced(BlockState state, @Nonnull World world, @Nonnull BlockPos pos, @Nonnull BlockState newState, boolean isMoving) {
```
maxanier: While you are at it, this should probably be migrated to #onBlockHarvested like BedBlock does and also check if the blocks we are destroying are actually tent blocks

maxanier (commented):
Maybe the sleep position can be shifted up a little bit. But if you already tried that, and it messed up the first person view, it's ok.

src/main/java/de/teamlapen/vampirism/blocks/TentBlock.java
```diff
@@ -146,11 +174,25 @@ public Direction getBedDirection(BlockState state, IWorldReader world, BlockPos
...
+        });
+    }
+
+    @OnlyIn(Dist.CLIENT)
+    public boolean hasCustomBreakingProgress(BlockState state) {
+        return true;
+    }
+
+    @Override
+    public void onReplaced(BlockState oldState, @Nonnull World world, @Nonnull BlockPos blockPos, @Nonnull BlockState newState, boolean p_196243_5_) {
```
maxanier: Is this necessary? onBlockHarvested should take care of this I assume. At least BlockBed does not use onReplaced

maxanier:
Awesome :heart:

maxanier:
Please always test changes that involve anything sided on a server as well.
Also always start the sever at least once to ensure no missing class errors occur

PR #630, opened 2020-03-10
use LazyOptionals /w Optionals for CaptureEntities and TotemTileEntity

### LINKED ISSUE #629: [Bug] [1.14.4] Village make the game hangs, chunks can't load, Alchemy table crashing server, coffin not dropping, minor bugs...
## Versions
- Minecraft: 1.14.4                
- Forge: 28.2.0
- Vampirism: 1.6.0-beta.1                        

## Issue Description
Somehow, when reaching out a certain village (we were on a server) which has been object of conquer and reconquer between vampire and hunter faction, the chunks within the village don't load and after that the server throw a "Timed out" and doesn't respond at all to commands nor chat messages. I tried to reproduce the same in singleplayer, and achieved this: basically the same except there is not "Timed out", just enitites stop move, chunks don't load and when you try to exit the game, it hangs with a "Saving world" screen.

## Reproduce Steps
I uploaded the server world in a .zip format [here](http://www.mediafire.com/file/j8sukk4xf4dm15p/vampirismoacab-world.zip/file)
Just /tp 300 80 300 and this happens
If you /tp 250 80 250 and you can see the chunks not loading but game is still "running"
Before all of this, the world filesize was a lot smaller, like before it was 25MB and after that the aternos server says 99MB. The village once it could be loaded without problem, but it has been attacked and reattacked since it was getting visited by me (vampire) and by my friend (hunter) often in the same hour.

## Additional Information
These are the final lines of the minecraft console when tipping myself to 200 100 200 (I'm using Tlauncher to launch my minecraft):
> [m[32m[09:17:56] [Server thread/INFO] [minecraft/MinecraftServer]: [budapest_god: Teleported budapest_god to 200.5, 100.0, 200.5]
> [m[32m[09:17:56] [Client thread/INFO] [minecraft/NewChatGui]: [CHAT] Teleported budapest_god to 200.5, 100.0, 200.5
> [m
**Here I am!**
[VersionManager] Refreshing versions locally...
[VersionManager] Versions has been refreshed (65 ms)
[Launcher] Minecraft closed with exit code: 1
[Launcher] Launcher exited.

I'm like, yo wtf, "Here I am"? Like the string before it's incomplete and then that message? This just seems like some Herobrine trolling

Cheaterpaul:
Replace the hard reference of the TotemTileEntity in CaptureEntities with LazyOptionals
 (@maxanier forge/vanilla solves the LazyOptionals null problematic with LazyOptionals of Optionals)
#629

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
+++ b/src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
@@ -61,6 +61,7 @@ import net.minecraft.world.gen.feature.structure.Structures;
 import net.minecraft.world.server.ServerWorld;
 import net.minecraftforge.api.distmarker.Dist;
 import net.minecraftforge.api.distmarker.OnlyIn;
+import net.minecraftforge.common.util.LazyOptional;
 import net.minecraftforge.registries.ForgeRegistries;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
@@ -153,8 +154,9 @@ public class TotemTileEntity extends TileEntity implements ITickableTileEntity {
         return new TranslationTextComponent("command.vampirism.test.village.success", faction.getName());
     }
 
-    public static VillageAttributes getVillageAttributes(TotemTileEntity totem) {
-        return new VillageAttributes(totem);
+    @Nullable
+    public static VillageAttributes getVillageAttributes(@Nullable TotemTileEntity totem) {
+        return totem == null ? null :new VillageAttributes(totem);
     }
 
     //block attributes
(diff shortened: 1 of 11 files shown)


src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
```diff
@@ -622,14 +624,14 @@ private void setCapturingFaction(@Nullable IFaction faction) {
...
+    @Nonnull
+    public AxisAlignedBB getVillageArea() {
+        return this.villageArea == null ? this.village != null ?this.villageArea = AxisAlignedBB.func_216363_a(this.village.getBoundingBox()) :new AxisAlignedBB(0,0,0,0,0,0): this.villageArea;
     }
 
-    private @Nonnull
-    AxisAlignedBB getVillageAreaReduced() {
-        return this.villageAreaReduced == null ? this.villageAreaReduced = AxisAlignedBB.func_216363_a(this.village.getBoundingBox()).grow(-30, -10, -30) : this.villageAreaReduced;
+    @Nonnull
+    private AxisAlignedBB getVillageAreaReduced() {
```
Cheaterpaul, line 633: it is possible that the entity ai task was updated before the world was even loaded ?! therefore a check

maxanier:
I am still not entirely happy with the design, not totally sure why though :see_no_evil: 
We still have a somewhat hard reference from the entity to the tile. So the entity will keep the tile alive even if the chunk has been unloaded. There does not appear to be a way to reset the entities VillageAttributes
However, I still don't know how to solve this. As the tile cannot know about the lazyOptional it cannot invalidate it.

I will probably merge this tonight anyway

PR #631, opened 2020-03-10
use LazyOptionals /w Optionals for CaptureEntities and TotemTileEntity

### LINKED ISSUE #629: [Bug] [1.14.4] Village make the game hangs, chunks can't load, Alchemy table crashing server, coffin not dropping, minor bugs...

Cheaterpaul:
this way should be total safe, but the solution is not really nice.

removed all hardreference to a synced LazyOptional<BlockPos> which the Tile invalidates on remove

#630
#629


src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
```diff
@@ -987,6 +1028,8 @@ public BlockPos getPosition() {
         }
 
         public boolean shouldForceTargets() {
+            totem.onChunkUnloaded();
+            totem.world.isAreaLoaded(pos, (int)villageArea.getAverageEdgeLength());
```
maxanier: Is this intended?

src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
```diff
@@ -153,8 +161,17 @@ public static ITextComponent forceFactionCommand(IFaction faction, ServerPlayerE
         return new TranslationTextComponent("command.vampirism.test.village.success", faction.getName());
     }
 
-    public static VillageAttributes getVillageAttributes(TotemTileEntity totem) {
-        return new VillageAttributes(totem);
+    @Nonnull
+    public static LazyOptional<Optional<BlockPos>> getVillageBlockPosOpt(Dimension dim, BlockPos pos){
+        LazyOptional<Optional<BlockPos>> opt = villagePosition.get(dim, pos);
```
maxanier: `villagePositition` does not appear to be filled anywhere

maxanier:
Looks quite advanced now :=)

But I am still not entirely comfortable with merging this.
Firstly it is very complex and requires multiple caches, just for a single instance where the tile is actually accessed (`shouldForceTargets`)

Secondly, that access has become rather costly now. For every capture entity this is retrieved every AI tick for every entity in the bounding box.

maxanier:
Looks like less overkill now :)
got to be properly tested though. Also still not sure about the AI stuff, still seems to be rather expensive. Maybe the lazyoptional could be cached outside the predicate since it will be invalidated by the tile on removal anway

PR #639, opened 2020-03-14
[1.14] fix Exception for entity tick while world is not loaded yet

Cheaterpaul:
removes dependency from VillageAttributes -> TotemTileEntity for all ICaptureEntities

#629

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
+++ b/src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
@@ -2,6 +2,7 @@ package de.teamlapen.vampirism.tileentity;
 
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
+
 import de.teamlapen.lib.lib.util.UtilLib;
 import de.teamlapen.vampirism.api.VReference;
 import de.teamlapen.vampirism.api.VampirismAPI;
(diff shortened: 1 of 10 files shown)


src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
```diff
@@ -947,32 +960,32 @@ public AxisAlignedBB getRenderBoundingBox() {
...
+        private final IFaction defendingFaction;
+        @Nullable
+        private final IFaction attackingFaction;
         private final AxisAlignedBB villageArea;
-        private final TotemTileEntity totem;
         private final BlockPos pos;
+        private final boolean shouldForceTargets;
 
-        private VillageAttributes(TotemTileEntity totem) {
+        private CaptureInfo(TotemTileEntity totem) {
```
maxanier, line 956: I think for things like this one usually uses a build/create method rather. But I don't really know why.
There is also absolutely no need to change this, just a note.

src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
```diff
@@ -450,19 +462,20 @@ public void tick() {
                         IFaction faction = VampirismAPI.factionRegistry().getFaction(entity);
                         if (faction == null) continue;
                         if (entity instanceof ICaptureIgnore) continue;
+                        CaptureInfo captureInfo = new CaptureInfo(this);
```
maxanier: Does this have to be created for every single entity in the area or can we use a single instance for all?

PR #651, opened 2020-03-28
fix to many villager in village

Cheaterpaul:
change from spawning villager if there are empty point of interests to spawn villager if there are enough beds


maxanier:
Maybe expand the check bb to make sure we count villagers that have run away a little bit.
Also some hard limit (math.min(beds,30)) or so.

PR #681, opened 2020-05-06
1.15 datagen

Cheaterpaul:
create datagenerator for data/assets resources


src/main/java/de/teamlapen/vampirism/core/ModTags.java
```diff
@@ -57,6 +57,19 @@
         }
     }
 
+    public static class Fluids {
+        public static final Tag<Fluid> BLOOD = tag("blood");
```
maxanier: For balacing reasons we probably don't want to use blood here. Other mods use blood in a very different way and if they choose to use the same tag, it would break balancing badly. So either "vampirism_blood" or not tag

Cheaterpaul:
f33e1a8  may be interesting

Cheaterpaul:
guide book recipe still failing while the item does not exist

PR #690, opened 2020-05-14
1.14 appearancescreen

### LINKED ISSUE #684: [Feature] Playermodel customization GUI
Currently, the only way to change the appearance of your fangs & eyes (as a Vampire obviously) is to use `/vampirism fang <0-6>` and `/vampirism eye <0-15>`. It's working, but not really an ideal way and players often don't know about new styles. So, maybe a customization GUI would be a great addition and could perhaps be added to the skill GUI. Here's a concept art:
gui1
gui2
gui3

Cheaterpaul:
add appearance screen for vampire

openable from skill menu

close  #684


maxanier:
Looks awesome.
Couple notes:
- The player model appears to be rendered a little bit outside of the GUI maybe the position can be adjusted
- It seems like the default eye is now no overlay and the fangs are white by default. The default look should remain the same
- The action reorder button hidden inside the appearance menu is a little bit weird. I know the space is limited in the skill menu. Maybe "change appearance" could be replaced by a small image based button showing just a image with a player face with fangs and eyes (e.g. my avatar), which is only displayed for vampires. Then the action button should fit in the skill screen again

maxanier:
As this is a "bigger" change I suggest we schedule this for the next major version 1.7.
Will create a branch for this eventually once we collected a few PRs

Cheaterpaul:
new are 6bf2c99, 2ffed89, 8a4b5e5

Cheaterpaul:
merged into #698

PR #696, opened 2020-05-18
[1.14/1.15] Totems consider only PointOfInterest instead of Structures

### LINKED ISSUE #694: [Feature] Create villages
Add a way to convert custom built villages into proper ones that work with the totem, if this is not too complex.

Maybe cheat only

Cheaterpaul:
transfoms the totems to only consider PointOfInterest for villages.

things to consider/existing issues:
  - destroyed PointOfInterests not removed from TotemTileEntity#vampireVillages
  - when should a collection of PointofInterests be considered a village
  - update BoundingBox if a PointofInterest has been added
  - when 2 villages collide update vampirevillage map
  - when 2 villages collide when one totem wins update the other entities

- remove https://github.com/Cheaterpaul/Vampirism/blob/f34347cb72bab9bb0c8e83af40aa16ebd53f10b8/src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java#L184

close #694

### DIFF
--- a/src/lib/java/de/teamlapen/lib/lib/util/UtilLib.java
+++ b/src/lib/java/de/teamlapen/lib/lib/util/UtilLib.java
@@ -706,6 +706,10 @@ public class UtilLib {
         return new MutableBoundingBox(array[0], array[1], array[2], array[3], array[4], array[5]);
     }
 
+    public static MutableBoundingBox AABBtoMB(AxisAlignedBB bb) {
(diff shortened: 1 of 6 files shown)


maxanier:
@Cheaterpaul Not sure how the village/point of interest stuff changed in 1.16. If 1.16 does things very differently, maybe we should only implement this feature for 1.16. Or alternatively, implement the change for 1.16 first and then check if it's possible to backport

PR #731, opened 2020-07-13
Taskmaster improvements

Cheaterpaul:
-  the taskmanager now has no general storage for all available tasks. Now they are saved based on a task board id (in this case the entityId of the taskmaster entity).
- the unique tasks are used by every task master
- the available tasks are randomly chosen from unlocked tasks for every task master.
  - if the tasks are not accepted until  gametime % 24000 = 0 the unaccepted task are replaced
  - maximum of the repeatable tasks per taskmaster are configured in the config (atm 4)
- different state of the tasks are rendered in different colors (maybe to much)
- task can be aborted (return to the not accepted state)


maxanier:
Nice, I noticed a few issues though:
- When relogging/loading after accepting a task the will be not accepted anymore (and progress is thereby lost)
- The tooltips render at a static position instead of relative to the mouse like e.g. in the inventory. This feels weird :D.
- There is no way to tell the progress of a entity kill task
- It might be cool to render the icons of completed subtasks half transparent or gray. Only if simple to implement (`push(); GLStateManager.color4f(0.3f,0.3f,0.3f,1f); render(); pop()`?
- The color help, but the color scheme feels a little bit off. Maybe gray/beige for not-accepted yet.  Light green/Light gold for uncomplete and green/gold for completed? Don't know

Cheaterpaul:
rendering the completed requirement icons in a different way is not possible, because i use the itemrenderer to render it. Only by copy paste the code this is possible. Other idea would be to dont render it at all

PR #736, opened 2020-07-17
Add JEI "recipe" renderer for task rewards

maxanier:
Therefore it is possible to see from JEI, that a certain item can be obtained as task reward.

Unlock requirements are also rendered.


Cheaterpaul (approved):
Looks fine, but i would render the prerequisites even if there are none (eg. with "none").

Additionally you could add an label for the applicant faction ("all" if no faction) and the task name in the top left corner (maybe even the requirements). To show all relevant information and have a bigger difference between two tasks that have the same itemstack as reward

PR #750, opened 2020-07-30
make village worldgen configs matter

Cheaterpaul:
The village structures must be added after the serverconfig is loaded (eg. FMLServerIsAboutToStart don't know if there is a better place). knowing this they must removed/re added when switching between world (client side). Should not make any difference on the server.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ org.gradle.daemon=false
 org.gradle.jvmargs=-Xmx3G
 #29.07.14 17:00
 minecraft_version=1.16.1
-forge_version=32.0.86
+forge_version=32.0.98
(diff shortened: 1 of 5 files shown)


maxanier:
With this no tents are generated at all? Do I need to add something beforehand?

I have no idea how worldgen works nowadays, so I can't really review this. If you tested it and considered all possible scenarios, that should be enough.

maxanier:
NVM 8825cf13d98dd85799bf31976e31f9c2f2648b40 broke tent generation I think

maxanier:
Nevermind, apparently I just had bad very luck. Tents seem too generate normally

maxanier:
Oh, derp, I pushed to the wrong branch. But now its on the right one anyway ^^

PR #753, opened 2020-08-10
[1.14] lets Totems cosider all villages as village and not only generated ones

Cheaterpaul:
make the totem PointOfInterest based

- [x] handle village colliding
  - totems compete over PointOfInterest (POI)
  - totem with 0 POI gets destroyed if in range of another totem 
  - villages do not merge when:
     - one is being captured
     - have different faction
  - village can get a POI if
     - village is natural generated
     - village is bigger
- [x] max village radius
  - default 100, configurable in balance config (maybe remove)
- [x] when should a village be considered a village
  - min 4 beds
  - min 2 workstations
  - min 4 villager
  - on shift click shows whats missing
- [x] POIs cant be destroyed like the totem blocks
- [x] add info for player if the village is too small
- [x] add javadoc
- [x] test

- [x] retest everything
  - [x] advanced hunter after capture command are bugging (no name, wrong hat)


src/main/java/de/teamlapen/vampirism/tileentity/TotemTileEntity.java
```diff
@@ -75,84 +74,41 @@
...
-     * @return {@code null} if no totem exists
+     * this method has no use anymore, because the village now persists of a collection of {@link PointOfInterest} and not based on a structure start
+     *
+     * @throws Exception
      */
+    @Deprecated
     public static @Nullable
-    BlockPos getTotemPosition(StructureStart structure) {
-        return totemPositions.get(structure);
+    BlockPos getTotemPosition(StructureStart structure) throws Exception { //TODO 1.16 remove
```
Cheaterpaul: @maxanier this method will not work with the new system. Should i be kept like this, or can it be removed or return null?

maxanier: Is there a replacement for the current use cases?
Since it is not exposed via an old API, I don't see a reason why we should keep it

Cheaterpaul: yes in TotemHandler with param `Collection<PointOfInterest>`.
Then i remove it

src/main/java/de/teamlapen/vampirism/player/ModPlayerEventHandler.java
```diff
@@ -345,4 +355,47 @@ public void eyeHeight(EntityEvent.EyeHeight event) {
             }
         }
     }
+
+    @SubscribeEvent
+    public void blockDestroyed(BlockEvent.BreakEvent event) {
+        if (event.getPlayer().isCreative()) return;
+        Set<BlockPos> positions = new HashSet<>();
+        BlockPos totemPos = TotemHelper.getTotemPosition(event.getPos());
+        Block block = event.getState().getBlock();
```
maxanier, line 366: Add some general comments roughly explaining stuff

PR #781, opened 2020-10-08
Update en_us.json

rabbit071:


maxanier:
Hey, thank you for the translations. Usually they are handled via our Crowdin page, but I can upload the file there.
There were many formatting errors which I had to fix.

Is this simplified or traditional Chinese?

maxanier:
Uploaded to crowdin as traditional Chinese

PR #795, opened 2020-10-26
better canBlockSeeSun testing

### LINKED ISSUE #779: [Bug] Farmland doesn't protect against sunlight
## Versions
- Minecraft: 1.15.2
- Forge: 3.1.2.36
- Vampirism: 1.7.0.beta.1
- ModPack: the self authored one I mentioned in my previous two issues (i.e. [this pack](https://www.curseforge.com/minecraft/modpacks/puncher-pack-6-golden-carrot-edition))

## Issue Description
Farmland doesn't protect against sunlight, i.e. if you stand under farmland during the day as a vampire and have no other blocks between you and the sky, you get the yellow glare and any other weaknesses corresponding to your level.

## Reproduce Steps
hoe some minecraft:dirt
stand under it
/vampirism level vampirism:vampire 4
/time set day
die!

Cheaterpaul:
add a Blocks#isSolid check for the seeSky test.

1. test for fluid
2. for solid
    - all solid blocks do not pass light, except for fences and fence doors ? so test if top or bottom is solid
3. test for opacity
    - leaves are not solid but do not let the sun pass

close #779

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/util/Helper.java
+++ b/src/main/java/de/teamlapen/vampirism/util/Helper.java
@@ -22,6 +22,7 @@ import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.entity.player.ServerPlayerEntity;
 import net.minecraft.item.ItemStack;
 import net.minecraft.profiler.IProfiler;
+import net.minecraft.util.Direction;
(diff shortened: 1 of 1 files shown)


Twisted-Code:
nice! (I'm the one that reported the issue with farmland)

PR #797, opened 2020-10-28
accept even non dye items as coffin coloring

### LINKED ISSUE #796: Use tags for coffin coloring dye
Make it possible to dye coffins, namely I need pink one.

No, it makes it more attractive to female players.

## Can you provide additional resources (etc. textures or models)
Not right away, but if you tell me how/give me input, I could play with GIMP or something.

Cheaterpaul:
close #796

### DIFF
--- a/src/lib/java/de/teamlapen/lib/lib/util/UtilLib.java
+++ b/src/lib/java/de/teamlapen/lib/lib/util/UtilLib.java
@@ -13,6 +13,8 @@ import net.minecraft.entity.ai.goal.GoalSelector;
 import net.minecraft.entity.ai.goal.PrioritizedGoal;
 import net.minecraft.entity.player.PlayerEntity;
 import net.minecraft.entity.player.ServerPlayerEntity;
+import net.minecraft.item.DyeColor;
+import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.CompoundNBT;
 import net.minecraft.particles.IParticleData;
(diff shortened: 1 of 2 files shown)


src/lib/java/de/teamlapen/lib/lib/util/UtilLib.java
```diff
@@ -721,4 +724,13 @@ public static AxisAlignedBB MBtoAABB(@Nonnull MutableBoundingBox bb) {
         HUNDRED_EIGHTY,
         TWO_HUNDRED_SEVENTY
     }
+
+    @Nullable
+    public static DyeColor getColorForItem(Item item) {
+        if (!Tags.Items.DYES.contains(item)) return null;
```
maxanier, line 730: Not sure if this is helpful for performance because it goes through to the entire set. The following line would go through 16 sets, but with about 1/16 of elements, so it shouldn't be much slower. Don't know, but probably does not matter anyway.

src/lib/java/de/teamlapen/lib/lib/util/UtilLib.java
```diff
@@ -721,4 +724,13 @@ public static AxisAlignedBB MBtoAABB(@Nonnull MutableBoundingBox bb) {
...
+        Optional<DyeColor> color = Arrays.stream(DyeColor.values()).filter(dye -> dye.getTag().contains(item)).findFirst();
+        if (color.isPresent()) return color.get();
```
maxanier, line 732: Why not stick to the optional here. In general I would say they are indeed more elegant and fit better to Forge style. But not a big difference anyway.

PR #798, opened 2020-10-29
add config command

### LINKED ISSUE #794: [Feature] better compatibility with Ice and Fire and other mods adding mobs
I'd like to see (haven't found yet, may be it's there, then I am sorry) some way to blacklist drinking blood of some entities, as it makes game heavily unbalanced - eg. When you kill dragon in Ice and Fire, you get dragon skull. Dragon skull is supposed to be "trophy", thus it is placable. When you place it, it behaves as an entity and when you "break" (in this case kill) it, you get the dragon skull back. Dragons are quite large animals, skulls might be bigger than player in some cases, so Vampirism mod calculates it as entity with a lot of blood and players then don't have to run around and "drink" animals, they just need to carry one skull with them and whenever thirsty, just place it and drink it. What's worse is that it is "renewable" because upon drinking everything, the skull just gets broken and drops itself as an item, so you can place it again.
I believe it might be issue with some other mods too, as they might add some entities who are eg. undead and you can already see from their texture that they contain 0 blood, like some skeletons - but as of now, those mobs are still "drinkable". It would be nice to have blacklist, or at least some basic compatibility layer with those mods.

Significantly

Cheaterpaul:
- blacklist iceandfire skulls
- remove config logging

config command options for blacklisting entities for blood calculation and command options for sunscreen config blacklist/whitelist

close #794


src/main/resources/assets/vampirism/lang/en_us.json
```diff
@@ -697,8 +697,18 @@
   "command.vampirism.argument.action.notfound": "The given action %s does not exist",
   "command.vampirism.argument.faction.notfound": "The given faction %s does not exist",
   "command.vampirism.argument.faction.notplayable": "You can not play as %s",
+  "command.vampirism.biome.notFound": "The biome %s does not exists",
```
maxanier, line 700:
```suggestion
  "command.vampirism.biome.not_found": "The biome %s does not exist",
```
No camel case in lang files just in case.
Also no *s because its already in `does`

maxanier: No camel case in lang keys. Just in case.
Also no *s since its already in `does`

src/main/resources/assets/vampirism/lang/en_us.json
```diff
@@ -697,8 +697,18 @@
...
   "command.vampirism.base.gender.success": "Successfully set gender for lord title display",
   "command.vampirism.base.lord.successful": "%s is now a %s lord level %s",
+  "command.vampirism.base.config.entity.blacklisted": "entity %s is now blacklisted",
```
maxanier: Start with uppercase. Also `is no longer...`

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
```diff
@@ -94,6 +97,57 @@ BiteableEntry calculate(@Nonnull CreatureEntity creature) {
...
 
+    /**
+     * checks if the entity type is blacklisted for blood calculation
+     * does not check if the entity is blacklisted in the server config ({@link #isEntityTypeBlacklisted(EntityType)})
+     *
+     * @param type the entity type
+     * @returns weather the entity type is blacklisted or not
+     */
+    public boolean isEntityTypeBlacklistedByDefault(EntityType<?> type) {
+        if (!VampirismConfig.SERVER.autoCalculateEntityBlood.get()) return true;
```
maxanier: Probably want to move this check to the top of `calculate`. Not sure if there was a reason to have it after the blacklist#contains check beforehand

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
```diff
@@ -94,6 +97,57 @@ BiteableEntry calculate(@Nonnull CreatureEntity creature) {
...
+        if (!(type.getClassification() != EntityClassification.MONSTER && type.getClassification() != EntityClassification.WATER_CREATURE))
```
maxanier: why not == ||| ^^

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
```diff
@@ -94,6 +97,57 @@ BiteableEntry calculate(@Nonnull CreatureEntity creature) {
...
+        return false;
+    }
+
+    /**
+     * checks if the entity type is blacklisted for blood calculation
+     *
+     * @param type the entity type
+     * @returns weather the entity type is blacklisted or not
+     */
+    public boolean isEntityTypeBlacklisted(EntityType<?> type) {
```
maxanier: Not really sure if we really need four small methods here. Since the blacklist is specifically for blood calculations the methods won't likely be used elsewhere, so no need to differentiate. One or two just make things much more clear.

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
```diff
@@ -94,6 +97,57 @@ BiteableEntry calculate(@Nonnull CreatureEntity creature) {
...
+
+    /**
+     * checks if the creature entity is blacklisted
+     *
+     * @param creature the entity to check
+     * @returns weather the entity is blacklisted or not
+     */
+    private boolean isEntityBlacklisted(CreatureEntity creature) {
+        if (isEntityTypeBlacklisted(creature.getType())) return true;
+        if (!(creature instanceof AnimalEntity)) return true;
```
maxanier, line 108: Move instanceof checks to the top since they are likely much less costly than the method and probably have the only a slighlty smaller chance of returning true

src/main/java/de/teamlapen/vampirism/config/VampirismConfig.java
```diff
@@ -194,6 +162,10 @@ public static void onReload(final ModConfig.Reloading configEvent) {
             villageSeparation = builder.comment("Village centers will be at least N chunks apart. Must be smaller than distance").defineInRange("villageSeparation", 8, 1, 100);
             builder.pop();
 
+            builder.push("entities");
+            blacklistedBloodEntity = builder.comment("Blacklist entities from auto calculated blood values").defineList("blacklistedBloodEntity", Collections.emptyList(), string -> string instanceof String && ResourceLocation.isResouceNameValid(((String) string)));
```
maxanier: also entirely blacklistts their blood value right? (new creatures only)

maxanier (commented):
Awesome, this will likely improve the user comfort a lot.
Just some nitpicking.
Also maybe there could be some explanation/help/feedback for the commands. E.g. if you type `/vampirism config bloodvalues (blacklist)`

src/main/resources/assets/vampirism/lang/en_us.json
Cheaterpaul: fixed

src/main/resources/assets/vampirism/lang/en_us.json
Cheaterpaul: fixed

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
Cheaterpaul: fixed

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
Cheaterpaul: fixed

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
Cheaterpaul: squashed

src/main/java/de/teamlapen/vampirism/entity/converted/BiteableEntryManager.java
Cheaterpaul: moved

src/main/java/de/teamlapen/vampirism/config/VampirismConfig.java
Cheaterpaul: yes, renamed

PR #809, opened 2020-11-19
make aggressive event is not usable

Cheaterpaul:
- make the MakeAggressive event usable to spawn own entities
- delete old/deprecated code from to do's

i don't know why i labeled some of the to do's "1.17", but i don't see a reason not to remove them with the other stuff


maxanier:
I changed them to 1.17, because we already have a release for 1.16 so this is a breaking change. But meh

PR #841, opened 2021-01-26
Update zh_cn.json

zorc:
Update zh_cn.json


maxanier:
Thank you for these translations!
Vampirism's translations are managed via Crowdin: https://crowdin.com/project/vampirism
You should be able to upload your translation files there, if not, I can do when I have time.
Any translations on Crowdin will be automatically included in the next build

maxanier:
Ok, I have uploaded them to crowdin and they will be included in the next release

Brun333rp:
if you don't mind me sharing my opinion here, I think that it would be easier and more practical if the translations were made here on Github, rather than on Crowdin

PR #845, opened 2021-02-04
add modcompat for api dependent runnables

Cheaterpaul:
The Problem is, that the API is setup in the VampirismMod constructor but needed in Werewolves constructor (Mods are loaded randomly)

I can not call https://github.com/TeamLapen/Vampirism/blob/d85fdd78f35d194b47898c17ca4562053a3f7380/src/main/java/de/teamlapen/vampirism/data/recipebuilder/SkillNodeBuilder.java#L46 in gatherdata without registering the faction, but the modbus events are not fired in run data, so i cannot register the faction outside the constructor.

This exact issue can also be resolved by adding another `faction(Resourcelocation loc)` method, but there might be other places where the playable faction is needed before the setup event.

### DIFF
--- a/src/api/java/de/teamlapen/vampirism/api/VampirismAPI.java
+++ b/src/api/java/de/teamlapen/vampirism/api/VampirismAPI.java
@@ -21,6 +21,8 @@ import net.minecraftforge.common.capabilities.CapabilityInject;
 import net.minecraftforge.common.util.LazyOptional;
 
 import javax.annotation.Nonnull;
+import java.util.ArrayList;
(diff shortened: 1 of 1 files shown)


maxanier:
If required in the future, such feature could be added here: https://github.com/TeamLapen/Vampirism/blob/f7dec552ae991de89db409b36e8977c2aa923853/src/api/java/de/teamlapen/vampirism/api/VampirismAPI.java#L122-L121

Must make sure to avoid problems with concurrency though.

PR #847, opened 2021-02-06
Add structure separation to worlds that are dynamically created

maxanier:


maxanier:
TODO test and think about positive/negativelist

PR #860, opened 2021-02-19
Reset player and minion skills with ingame elements

### LINKED ISSUE #849: [Feature] Allow resetting minion stats

<!-- This may include suggesting crafting recipes, item names or places in skill tree if applicable ->>
E.g. in the GUI. But also taking away their levels

### LINKED ISSUE #853: [Feature] Add task to reset skills

Maybe it would be nice to add a task that resets the skillpoints (without loosing an level).

It would be more integrated into the game instead of being just a GUI button

Cheaterpaul:
close #853
close #849

Adds Potion of Oblivion that, when consumed, gives the player the oblivion effect resulting in forgetting the learned skill points one by one.

Also works for minions, by clicking the reset button in the minions skill screen when they have a potion of oblivion in their inventory.

- [x] add texture
- [x] add way to obtain the item
- [x] handle minion inventory shrink
- [x] disable unlocking skills while under the effect of the potion of oblivion


src/main/java/de/teamlapen/vampirism/core/ModTasks.java
```diff
@@ -87,6 +88,8 @@ public static void registerTasks(IForgeRegistry<Task> registry) {
         registry.register(TaskBuilder.builder().addRequirement("skeleton", EntityType.SKELETON, 10).setReward(new ItemStack(Items.CHAINMAIL_HELMET)).build("break_bones4"));
         registry.register(TaskBuilder.builder().unlockedBy(new LvlUnlocker(5)).addRequirement("obsidian", new ItemStack(Items.OBSIDIAN, 32)).addRequirement("diamond", new ItemStack(Items.DIAMOND, 1)).addRequirement("zombie", EntityType.ZOMBIE, 32).setReward(new ItemStack(ModBlocks.totem_top_crafted)).build("totem_top"));
 
+        registry.register(TaskBuilder.builder().addRequirement("poison", PotionUtils.addPotionToItemStack(new ItemStack(Items.POTION), Potions.POISON)).addRequirement("vampire_blood", new ItemStack(ModItems.vampire_blood_bottle)).setReward(new ItemStack(ModItems.oblivion_potion)).build("oblivion_potion1"));
```
Cheaterpaul: The potion is currently obtainable as task for 1 vampire blood & 1 potion of poison.
With two tasks to increase the probability.

src/lib/java/de/teamlapen/lib/lib/inventory/InventoryHelper.java
```diff
@@ -155,19 +155,55 @@ public static void addStackToSlotWithoutCheck(IInventory inv, int slot, ItemStac
...
+        int i = item.getCount();
+
+        for (int j = 0; j < inventory.getSizeInventory(); ++j) {
+            ItemStack itemstack = inventory.getStackInSlot(j);
+            if (itemstack.getItem().equals(item.getItem())) {
+                if (itemstack.getCount() >= i) {
+                    itemstack.shrink(i);
+                    return true;
+                } else {
+                    i += itemstack.getCount();
```
maxanier: Is this correct?

src/main/java/de/teamlapen/vampirism/core/ModTasks.java
maxanier: Can you obtain vampire blood as a vampire?

Issue #806, opened 2020-11-09
[Feature] Vampire armor replacement


maxanier:
An extensive suggestion to implement #287
By ♤SpadeIsACard♤,S_olace , spoopy special krab:

# Vampire armor idea/suggestions

How this would help balance the game

In the hunter class skills are chosen to help improve the weapons used because they have no magic, so vampires should have items that work off of the skills they choose since they only have magic. This would help give variety to how being vampire can be played and it&#39;s been stated before that you don&#39;t want vampires to have armor in the future so with the absence of vanilla defense this leaves room for possible cosmetics and vampire items that can go into those empty armor slots.

Summary of our suggestion

The item(s) that we&#39;re suggesting are essentially accessories that can go in armor slots to replace vanilla defence with customizing your stats, these items would act like armor in MMOs in the way that they boost different skills. We have four different kinds of accessories, each of which can go into an armor slot. They are then set into five types of rarity -common \&gt; uncommon \&gt; fine \&gt; rare \&gt; legendary- which determine how you can acquire them and what abilities they have. Common and uncommon being craftable and the rest are either drops or found in chests, higher in rarity, harder to find but greater abilities. Each in question could also have a small armor rating like the vampire cloak to give at the very least some protection.

# Rarity &amp; Item Types

**Common:** these would be craftable, using things such as iron, gold, etc. you would have to enchant these by either an enchantment table or a charging station. These would give small buffs such as speed, regen, low blood loss.

**Uncommon:** these would still be craftable but cost more to make, involving diamonds/gold/vampire book or something similar. These would still grant the same buffs, however, these buffs would be stronger.

**Fine:** these would be uncraftable, only found in any dungeon/ sold by vampire experts. These would provide flight changes to vampire abilities as a whole. Faster bat movement speed, garlic affects you less, etc.

**Rare:** these would be much rarer and only found in mineshafts, vampire dungeons, or dropped rarely from advanced vampires. These would give actual ability changes, such as turning into multiple bats for more health but slower movement and larger hitbox.

**Legendary:** These would only be found extremely rarely in vampire dungeons and rarely dropped off of level 5 barons. There would only be 5 variations (not 5 per world just 5 versions). These would also have the requirement of being a vampire lord to use (must have 1-5 minions to use).

**Common**

**Ankle bracelet:** these would be smaller ring-like objects that would be equipped to the boot section of one&#39;s inventory.

[...]

PR #908, opened 2021-04-03
Vampire natural armor

### LINKED ISSUE #287: No armor for vampires
Vampires and armor do not really fit together. Vampires are fast and agile creatures whereas MC armors are always clunky and heavy.

Vampires already gain a strong boost in health and strength, so they should be able to survive without armor. Therefore I am currently considering somehow preventing vampires to wear armor parts. 
Instead they could be able to wear special rings or charms which give other boosts. However, I am expecting quite some resistance against this change, so I am not sure yet.

### LINKED ISSUE #906: [Feature] Vampire cloaks feel useless
Vampire cloaks are outclassed by practically any chestplate in the game, so even the cost of one diamond feels not worth it. Overall, they feel like a kinda-cool-but-why feature. I heard something about being to upgrade them, but I couldn't find any indication of such a possibility existing.
Thus, I would suggest adding some kind of ability to improve a cloak - either by using it so it grows as you do or performing some kind of ritual. Upon levelling up the cloak may give more defence or some other buff like speed/jump boost/thirst rate reduction etc.

I think so. It would make a currently underpowered item into an actually useful one.

maxanier:
Vampires have a natural armor / tough skin. This grants them armor protection without wearing armor.
The armor value starts between leather and iron and scales up to diamond level at max vampire level.
When equipping armor anyway, the higher value (worn armor/natural armor) is applied. But the vampire speed and attack damage is reduced (except for leather armor and below).
This allows wearing armor from other mods, if they provided very strong armor that is required for gameplay.
Not sure about armor enchantments yet

#806
Closes #287
Closes #906


maxanier:
TODO
- [ ] Add enchant-able vampire clothing

Cheaterpaul:
Should the armor regeneration be cancelable by drinking milk? If not `IForgeEffect#getCurativeItems` must be overridden.

maxanier:
Enchant-able clothing can be added later

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -1212,6 +1226,81 @@ private void updateFeeding() {
...
+                armorAtt.removeModifier(NATURAL_ARMOR_UUID);
+                toughnessAtt.removeModifier(NATURAL_ARMOR_TOUGHNESS_UUID);
+            }
+            else{
+                AttributeModifier modArmor = armorAtt.getModifier(NATURAL_ARMOR_UUID);
+                AttributeModifier modToughness = toughnessAtt.getModifier(NATURAL_ARMOR_TOUGHNESS_UUID);
+                double naturalArmor = getNaturalArmorValue(lvl);
+                EffectInstance armorRegen = player.getActivePotionEffect(ModEffects.armor_regeneration);
+                double armorRegenerationMod = armorRegen == null ? 0 : armorRegen.getDuration() / ((double)VampirismConfig.BALANCE.vpNaturalArmorRegenDuration.get() * 20);
+                naturalArmor *= (1-0.5*armorRegenerationMod); //Modify natural armor between 50% and 100% depending on the armor regen state
```
Cheaterpaul: is 50% a little  too much? I would prefer a lower base value and a higher regeneration.

Cheaterpaul (approved):

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
maxanier: I thought iron might be a good starting point (if you are at level 14). But yeah, it should regenerate fairly quickly, so I changed it to 25% for now

PR #918, opened 2021-04-20
Don't modify zombie/skeleton ai. Instead cancel their attack target

maxanier:
TODO
- [ ] Test


maxanier:
This does not work unfortunately. We still want to allow e.g. revenge attacks, however, the revenge timer is reset after just 5 seconds. Previously the mob would continue attacking the player, but with this change it does not

PR #922, opened 2021-04-24
Taskinstance

Cheaterpaul:
Introduce task instances that represents a task with additional details

- add support for more configurations


maxanier:
General note: Maybe make sure shit doesn't blow up when tasks are removed

PR #925, opened 2021-04-28
Curevillager

### LINKED ISSUE #854: [Feature] Cure villagers

Instead of having to kill all vampire villagers when trying to capture a village it would be cooler, if one could cure them.
This could be either done immediately with a garlic injection or by giving them a new effect (like reverse sanguinare) by garlic injection or by a completely new way.

It could also give hunters some reward or be a task requirement
Maybe reduce prices after curing

Cheaterpaul:
close #854 

cure vampire villager with an infested apple while having the weakening effect


maxanier:
Looks good, though the general idea was to be able to cure any weak vampiric creature, so any `ConvertedCreatureEntity`, right?

I would also suggest to give the player the option to inject the zombie syringe (at a med chair) and just give them toxicant then (as their VP is too strong)

PR #926, opened 2021-04-30
mixin effect instance to prevent canceling of other effects

### LINKED ISSUE #895: Check/Rework the effect instance usage
For several features effect instances are added to a player and removed later on.
This means any pre-existing effects (and then merged) will be removed as well.
Maybe we can work with custom EffectInstance subclass. But we have to take care with regards to merging as well as save to nbt

Cheaterpaul:
close #895

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/mixin/MixinEffectInstance.java
+++ b/src/main/java/de/teamlapen/vampirism/mixin/MixinEffectInstance.java
@@ -1,0 +1,74 @@
+package de.teamlapen.vampirism.mixin;
+
+import de.teamlapen.vampirism.api.entity.effect.EffectInstanceWithSource;
+import net.minecraft.nbt.CompoundNBT;
+import net.minecraft.potion.Effect;
+import net.minecraft.potion.EffectInstance;
+import net.minecraft.util.ResourceLocation;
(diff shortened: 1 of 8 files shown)


src/main/java/de/teamlapen/vampirism/mixin/MixinEffectInstance.java
```diff
@@ -0,0 +1,74 @@
...
+        return this.hiddenEffects;
+    }
+
+    @Inject(method = "func_230117_a_(Lnet/minecraft/potion/EffectInstance;)V", at = @At("TAIL"))
+    private void copySource1(EffectInstance other, CallbackInfo ci){
+        this.source = ((EffectInstanceWithSource) other).getSource();
+    }
+
+    @Inject(method = "writeInternal(Lnet/minecraft/nbt/CompoundNBT;)V", at = @At("TAIL"))
+    private void writeInternale_vampirism(CompoundNBT nbt, CallbackInfo ci){
```
maxanier:
```suggestion
    private void writeInternal_vampirism(CompoundNBT nbt, CallbackInfo ci){
```

maxanier (commented):
Nice, but I can't figure out what `hiddenEffects` does and if it is even used

PR #931, opened 2021-05-13
introduce Faction Raids

### LINKED ISSUE #598: [Feature] Pillager Raids into Vampire Hunter Raids
It'd be cool that instead of Pillagers attacking your Villagers, that you could perhaps set off a similar thing with Vampire Hunters and Vampire Villagers, such as killing an advanced hunter or a group of wandering hunters similar to Pillagers. 

Not sure if it would make it more balanced but it would be more cool

## Can you provide additional resources (etc. textures or models)
Nnnno ;v;

Cheaterpaul:
close #598 

- village captures do not require player to stay in the village
- add new bossinfo bar
- add abort capture command
- add faction banners
- add badomen effects for each faction

Going into a villages with a bod omen effect (whether you are in a faction or not) a faction raid is triggered if the bad omen faction does not match the village faction.
BasicHunter/BasicVampire now have a small chance to spawn with a faction banner equipped. Upon death with a faction banner equipped the killing player receives the appropriate bad omen effect


src/main/java/de/teamlapen/vampirism/entity/hunter/BasicHunterEntity.java
```diff
@@ -474,6 +477,16 @@ public void writeAdditional(CompoundNBT nbt) {
         }
     }
 
+    @Override
+    public void onDeath(DamageSource cause) {
+        if (cause.getTrueSource() instanceof PlayerEntity && this.villageAttributes == null) {
+            if (ItemStack.areItemStacksEqual(this.getItemStackFromSlot(EquipmentSlotType.HEAD), HunterVillageData.createBanner())) {
```
maxanier: Is it possible to use `#getBanner` or even a new method `isBanner` here to avoid creating a new item stack when one is cached anyway

maxanier (commented):
I like the new boss bar :)
Not sure about the colors though. Having red for attackers and blue for defenders makes sense, but might cause confusion since usually vampires red/purple and hunters blue.

maxanier:
Regarding banner mobs:
So far I was unable to spawn a banner carrying hunter using a large array of tents. Not sure if this is working.

Also I'm not sure about the idea in general. Doesn't really make sense for a lone vampire to run around carrying a banner. But I don't have a better suggestion so far.

Cheaterpaul:
- bossbar color now uses the faction color
- hunters from tents can now have a banner

Cheaterpaul:
- random raids approximately every 2 hours, but with a minimum of 5 minutes since the last raid
- (extendible) strength ratio calculator with village event

PR #948, opened 2021-06-05
Check for null in LootingLevelEvent

vemerion:
Unfortunately, due to interactions with global loot modifiers, the damage source can be null in LootingLevelEvent. This fix avoids a potential null pointer exception by checking to make sure that the damage source is indeed not null.


maxanier:
Thank you for the PR.
Do you have any reference where this is actually null?
From the vanilla place vanilla checks the loot (`LivingEntity#spawnDrops`) the damage source cannot be null as it would already crash there.

vemerion:
Sure! Here is the issue on the Forge github repo which goes into more details: https://github.com/MinecraftForge/MinecraftForge/issues/7714

Cheaterpaul:
Yes `LootContext#getLootingModifier` can produce a null value

maxanier:
Thank you

PR #970, opened 2021-08-12
Update en_us.json

Brun333rp:
- Fixed several capitalization inconsistencies, mainly these on item names;
- Improved the names of certain advancements and capitalized them as well;
- Added descriptions to several advancements that lacked one;
- Changed the Crossbow name to "Basic Crossbow", to avoid confusion with vanilla crossbows;
- Changed "Cursed Earth" to "Cursed Soil", to avoid confusion with planet Earth.


Brun333rp:
I have a question, does the lacking descriptions (and even names) for advancements already have registered IDs? because if not, the changes I made regarding them won't work

Cheaterpaul:
The new translation keys are not used atm but i will add them.

Cheaterpaul:
Thanks for taking a look at the english translation and improving it.

@maxanier when adding new translations we should take a look at how vanilla names things. For example, almost all words in proper names start with a capital letter (blocks/items/entries ....)

PR #988, opened 2021-10-08
add pieces to rawTemplate field for RS compat and added structure void to vampirism's vanilla village nbt pieces

TelepathicGrunt:


maxanier:
Thank you for your contribution again. Vampirism's worldgen would probably much worse without your input.
The street trick sounds like a perfect solution for the hunter trainer house. I've put it on our todo list.

In general, you can probably tag Cheaterpaul instead of me on Discord regarding any worldgen / RS related topics.
I've long given up on trying to understand MC worldgen and dynamic registries, but Paul is successfully tackling these issues.

PR #1010, opened 2021-11-19
Active turning

### LINKED ISSUE #805: [Feature] Let vampires actively choose between biting/drinking/turning
In my opinion, vampires should be able to actively decide, whether to drink blood / bite an entity, or to turn it.

*(I suggested this on Discord a while ago, but it didn't quite work out. However, I've invested some time in refining the concept further.)*

---

#### In most modern stories/universes containing vampires, a transformation works like this:
1. The vampire drinks a little amount of blood from the human
2. The human needs to drink a bigger amount of blood from the vampire (most commonly the vampire bites his own wrist for this)
3. (Optional: The human must die, to be reborn as a vampire) *
4. The human transformed into a vampire and feels a rush of adrenaline and the extended senses for a few moments (equivalent to the saturation effect in MC?)

_* This would heavily change the way Vampirism does it currently and would lead to further questions, eg. "should the player loose their items on this particular death", etc., so I'd say this is **really** optional and can stay how it is currently (sanguinare effect)._

---

#### So, how could the transformation look like in Vampirism?
Currently there's one keybind to utilize the fangs. Pressing it will deal damage, holding it will (try) to consume blood. That's a pretty good solution. However, turning a human into a vampire should be different from just drinking their blood by holding the key.

The ability to turn someone could be an unlockable skill, rising the current maximum level from 14 to 15: 
tree_concept

Alternatively, if you don't want to make it it's own unlockable skill resulting in 15 levels, the ability could also be part of the current level 4, renaming it to "No longer a fledgling" and unlocking both bat mode and the ability to turn someone. *I really like the sound of that.* Players who didn't unlock the skill yet, should not be able to turn players, but randomly infect other entities, basically how it works currently. After unlocking the skill, they're no longer a fledgling and can actively control it now.

For the actual turning, I've been thinking of a (probably?) rather easy solution. Point at the entity you want to turn, open your action menu and select the turn skill.

---

#### About other entities...
Currently, mobs either die or turn once their blood is depleted. That should be changed as well:
- Pressing the (current) key: deal damage
- Holding the (current) key: drink blood
- Using the new skill: turn the entity with 100% success rate

If an entity suffers from blood loss (aka. is drained dry), it should just die, without having the chance to turn, just as players do.

---

#### To sum up...
- The current key press/hold mechanic should definitely stay the same:
     - Pressing: deal damage
     - Holding: drink blood
- The new skill to turn entities should be accessible from the action menu and should support being bound to a key using ``/vampirism bind-action``
- Vampire NPCs, the fang item & a sanguinare syringe should still be able to turn a player like they do now
- Entities should follow the same rules as players, when being the subject to being bitten (damage/blood loss/turn, depends on what the vampire has chosen)

maxanier:
Closes #805 

TODO:
- [x] feedback if you can turn the entity you are looking at
- [x] feedback why infection might have failed
- [x] better feedback when infection succeeded


maxanier:
One solution would be to have the action be a lasting one that is toggled to decide whether to feed or infect an entity when biting. Would prefer a different solution though

src/main/java/de/teamlapen/vampirism/player/vampire/VampirePlayer.java
```diff
@@ -601,9 +622,6 @@ public int onBite(IVampire biter) {
             int sucked = (int) Math.ceil((amt * perc));
             player.getFoodData().setFoodLevel(amt - sucked);
             player.causeFoodExhaustion(1000F);
-            if (!player.hasEffect(ModEffects.sanguinare) && Helper.canTurnPlayer(biter, player) && Helper.canBecomeVampire(player)) {
-                if (!player.isCreative()) SanguinareEffect.addRandom(player, true);
-            }
```
Cheaterpaul, line 606: @maxanier how are vampire player infected by mobs or should they not be able to do this? Without this there is no way they can do it. VampirePlayer#tryInfect is only called from the InfectAction

maxanier: Mhh, can't remember this, so it is probably not really intentional.
Maybe we can add an according entity action. For now I just added the according tryInfect hook to the vampire base entity attack

PR #1016, opened 2021-11-28
Decorative blocks II

maxanier:
Work in progress
### TODO
- [x] Multi-block-model block handling
- [x] Sitting functionality https://github.com/bl4ckscor3/Sit/issues/19
- Other blocks

### Final checklist
- [x] Block durability
- [x] Crafting recipes
- [x] Particle texture
- [x] Lang keys


src/main/java/de/teamlapen/vampirism/blocks/ThroneBlock.java
```diff
@@ -0,0 +1,31 @@
...
+
+public class ThroneBlock extends VampirismSplitBlock {
+    public ThroneBlock() {
+        super("throne", AbstractBlock.Properties.of(Material.WOOD).strength(2, 3), BlockVoxelshapes.throneBottom, BlockVoxelshapes.throneTop, true);
+        markDecorativeBlock();
+    }
+
+    @Override
+    public ActionResultType use(BlockState state, World world, BlockPos pos, PlayerEntity player, Hand p_225533_5_, BlockRayTraceResult traceResult) {
+        if (state.getValue(VampirismSplitBlock.PART) == Part.MAIN && traceResult.getDirection() == Direction.UP) {
```
Cheaterpaul: While testing i noticed that i does not feel right that you can only access the throne from the top. I would suggest allowing the throne to be accessible from the front direction on the bottom and top block.

Cheaterpaul (approved):

PR #1028, opened 2021-12-18
rework vampire forest

Cheaterpaul:
Reworks the vampire forest

- forest is dark themed with dark red as contrast color
- It has two nearly identical types of trees where one of them pulls non vampire entities in and drains their food level
  - implemented by bark that evolves around the log block
  - two new wood types

- [x] check if there is something left to do with the wood items/blocks
- [x] remove blood/hunger when sucked to the trees 
- [x] only suck non vampire
- [x] ensure backwards compatibility


maxanier:
The trees are amazing!! :D

PR #1030, opened 2022-01-01
rebuild skillscreen

Cheaterpaul:
Redesign the skillscreen to look like the newer advancement screen.
Build more expandable

- [x] fix the tool-tip rendering when zooming
- [x] move used vanilla assets to own files
- [x] skill tree centering
- [x] add "unlock parent first" label
- [x] render playerlevel in vampirism menu


maxanier:
Looks cool, but you will have fun redoing this for 1.17+ 

Any reasons for keeping the old SkillsScreen class? Also it is still referenced in ClientProxy

The old screen renders the current lord level instead of "Skills" on the top left. If we remove this here, it should at least be rendered somewhere else

Cheaterpaul:
SkillsScreen is not removed to avoid breaking changes

PR #1031, opened 2022-01-02
add lord skills

Cheaterpaul:
This PR adds an infrastructure to have different skill classes that can work separately in different trees.
Additionally to the infrastructure it provides the LORD skill class and skill trees for vampires and hunters.

Lord level now provide 1 Skill point per level which  can be used to unlock lord skills or normal skills.

vampire lord skills

Current lord skills:
- Minion stats increase (increases minions stats by 1.2)
- Lord speed (provides movement speed effect to nearby entities of the same faction)
- Lord attack speed (provides haste effect to nearby entities if the same faction)
- Collect (unlocks the collect blood / collect resources minion task)
- Recovery (minions recover in 0.8 of the time)
(each faction needs its own skill since a skill needs a faction)

Note:  
The Select action screen might have to much actions as Vampire, this may need a restructure.


Cheaterpaul:
Lord skill ideas:
- buff minions
- actions that increase stats of nearly faction entities

PR #1037, opened 2022-01-12
expose IIngameOverlays in api class

Cheaterpaul:
create IIngameOverlay classes and expose them in VIngameOverlays


src/main/java/de/teamlapen/vampirism/proxy/ClientProxy.java
```diff
@@ -239,6 +244,12 @@ private void registerSubscriptions() {
         MinecraftForge.EVENT_BUS.register(new ScreenEventHandler());
     }
 
+    @Override
+    public void setupAPIClient() {
+        VIngameOverlays.BLOOD_BAR_ELEMENT = this.bossInfoOverlay = new CustomBossEventOverlay();
```
maxanier: ?

PR #1058, opened 2022-03-13
1.18.2 Update

maxanier:
TODO:
- [x] Find a way to ensure specific village buildings arre generated only once
- [x] Modify village separation. (Maybe pass on that)
- [x] Verify hunter camp generation and set up respective tag
- [x] hunter_trainer/church/totem village buildingss
- [x] Update world gen related API
- [x] vampire forest + surface rules

Also test everything

Later:
- [ ] Make village modification better compatible with data pack added things
- [x] Readd RepurposedStructure village compat again. For piece count limit see [https://github.com/TelepathicGrunt/RepurposedStructures/tree/1ffe99074dcdecee9e2d5e8cc90520644a7ee1cd/src/main/resources/data/repurposed_structures/rs_pieces_spawn_counts]


maxanier:
Village gen is somewhat weird in 1.18.2.
- Even in Vanilla MC there are often empty streets that look like it failed to generate houses there.
- For example the desert hunter trainer house generates only very seldom despite being in the pool a hundred times. Somehow it fails to generate, but this also applies for the vanilla `desert_mansion_1` building it is based on. Apparently, that building type does not connect well (`JigsawBlock#canAttach`).
- Same as above applies for the desert temple (although it has a much lower probability)

maxanier:
I decided to refractor the biome/feature/gen stuff a bit.
ModBiomes and ModFeatures now only contain things directly related to ForgeRegistries.
Anything directly registered to the new BuiltinRegs or similar is now in `VampirismBiomes` and `VampirismFeatures` respectively. Things here might eventually be moved to data, if Mojang continues its current path.
Modifications to the "default" overworld are in `OverworldModifications` and `VanillaStructureModifications`
What do you think @Cheaterpaul

Cheaterpaul:
sounds about right

Cheaterpaul:
should the repurposed structure compat be completely done by datapacks? I.a. adding the jigsawpieces also via the datapack.

maxanier:
If that is possible, yeah. Main issue might be, that it shouldn't cause any issues if Repurposed Structures is not installed

Cheaterpaul:
both datapack folder are only loaded by repurposed structures

PR #1070, opened 2022-04-17
Forest Boss

maxanier:


maxanier:
Implementation details:

- Indestructible base block
- Requires graph search to find all structure blocks (implementing interface) from any given point in structure
- Central block fully enclosed by structure
  - Loot table chest like
- Weak points
  - Convert to block entity upon damage
  - Require continuous damaging for full destroying
  - Trigger defensive forces upon damaging
  - Slowly heal if not damaged
  - Convert to destroyed weakpoint block after destruction
  - Animated texture (while active)
  - Upon full destruction trigger structure integrity check based on graph search. If no further intact weakpoints are found, trigger structure destruction process
  - Potentially, activating one weakpoint could lock all for some time
  - Damage to block entity is collected via dummy entity
    - BE spawns entity with slightly larger BB
    - remembers ID to check if existent
    - Dummy entity forwards damage to BE
    - Dummy and dummy id in BE is not saved to disk, but respawned upon load
 - Structure destruction process (all block dissolve into air revealing central loot "chest")
 - WorldSavedData `ForestBossManager`
   - Not sure if actually needed
   - Fight id is derived from block pos of central block, weakpoint tile entities can search for it once, then cache id
 - Defense
   - Weakpoint BE continuously spawns vampires or new mob type nearby until destroyed or healed
   - Occasionally, new blocks are placed nearby (on ground/structure) that drain blood from nearby players, but can be destroyed with normal tools (axe)
 - Mob type
   - Spawned in ground
   - While in ground, does not suffocate, and slowly rise upwards
   - When above ground, reset collision box, and attack normally
 - Specific player action weaknesses
   - Freeze action temporarily "freezes" weakpoint defense (need to visuallize somehow)
 
# Motivation for fight
## General
- faction specific achievement
- unique decoration block or armor piece
## Vampire
- Crafting ingredient for fog generator (which may also function as saturation beacon)
- enhance vampire cloth
## Hunter
- Maybe item that can replace vampire blood without being consumed (if we add more use for vampire blood)

maxanier:
Moved to issue #1118

PR #1075, opened 2022-04-25
Update coffin model

### LINKED ISSUE #311: Upright coffin
Probably rather time consuming to implement. Not sure if worth the effort. But would look cool

Cheaterpaul:
- updates the coffin model to a new model by Йода
- adds dyed coffin block rather than a dye-able coffin tile entity
- adds vertical coffins close #311

Questions:
- [x] should villager be able to sleep in coffins? maybe later
- [x] should the cross be removed from the texture?

Todos:
- [x] crafting recipes?
- [x] fix server-client sync issues (shift right click, shape)
- [x] more testing


lusifer173:
Wanna mention, this isnt a coffin. It's a casket, there is a difference. Mainly in the shape, caskets being rectangles. _It really needs to be renamed. _

PR #1086, opened 2022-05-25
add dark spruce and cursed spruce boats

Cheaterpaul:


maxanier:
That's a lot of code to maintain for such a minor feature

maxanier:
I don't think not having cursed plank boats is that big a deal, but ok. Going to merge it this time

PR #1087, opened 2022-05-26
(1.18) Switch registration to use DeferredRegisters

lukebemish:
This pull request replaces the use of registry events and ObjectHolders with deferred registers. This has several benefits: firstly, it is more future-proof with regard to possible registry changes in 1.19, and second, it makes it substantially harder to accidentally pass null values before an object is registered, and the RegistryObject can be passed instead.

Changes:
- With the exception of the registry-replacement of night vision, every use of registry events and object holders for registering and holding items was replaced with deferred registers and registry holders.
- With the same exception, all uses of `setRegitryName` were removed.
- Spawn eggs and buckets are now constructed with Forge's methods that use suppliers.
- Item-based rewards in Tasks now use suppliers
- RefinementSets now hold sets of IRefinement RegistryObjects
- Certain registration-related code (such as setting up village trades) moved to common setup instead

This registration change should be less dependent on class-load order and on registration order - the old system, for instance, static-inited some entity types, which this should avoid.

*Sponsored by the [Modding Inquisition](https://github.com/TheModdingInquisition)*


maxanier:
Wow, that was a lot of work and some long overdue cleanup to the grown code.
The objectholders are still from the time before DeferredRegister existed, but I still somehow prefer them and never had any real issues with them. However, it seems that Forge is pushing towards DeferredRegister, so this is a reasonable change.

The issue is, that we are maintaining both 1.18 and 1.16 at the moment, but main development is currently focused on 1.16 (hence it is the main branch). Therefore, we tried to minimize the structural changes to 1.18 to simplify porting the new features to 1.18. Any structural changes per scheduled for the transition from 1.16/1.18 to 1.18/1.19.
This PR changes so many things, which means that cherry picking commits from the 1.16 to 1.18 will require many manual fixes.

If you plan such massive PRs, it would probably be reasonable to discuss them with the original authors first (even though your slogan is "Nobody expects the Modding Inquisition!" which is perfectly applicable here :D)

I will have to think about this. Not sure yet how to make use of this code cleanup and appreciate your work without increasing my workload massively. (I barely have time for anything at this time).

Matyrobbrt:
We could apply the same registry fixes to the 1.16 version if necessary.

lukebemish:
Yeah, it doesn't look like switching 1.16 to deferred registers as well would be too bad. We'll submit another PR once we've got that done.

maxanier:
I think switching 1.16 to deferred registers is not really necessary. The current code works just fine and won't be further updated after the next major update (which we are currently working on).
Hence, I propose the following. I will leave this PR open until we have ported that update to 1.18. Then we can rebase/update this PR. I think that requires less work than either applying the same changes to 1.16 or porting the the features to 1.18 after merging this. Does that sound reasonable?

Matyrobbrt:
Switching to 1.16 DR is almost done, however, we have finished blocks and items already. I think it is good if the last release of 1.16 uses best practices.

maxanier:
Ok, that's fine with me as well

lukebemish:
Hmm... Not quite sure why the checks are failing here

lukebemish:
I can take a look tonight

maxanier:
:raised_hands: 

Do you plan to do any other (larger) PRs to Vampirism? Just asking, because I plan to branch of 1.19 soonish

Matyrobbrt:
Yes, we are planning to fix the vampire spawning since Vampirism doesn't fire any EntityJoinWorld events when spawning entities, rendering mods like InControl (which limit mob spawns (based on a few rules), using that event in order to reduce lag when there are many players) useless

Though, we will wait until the 1.19 branch is created, if 1.19 becomes the focus

maxanier:
Isn't that event automatically fired when using `addFreshEntity`?
Anway, if you expect many changed files, probably best to just do it for 1.16. We will port it to 1.18/1.19.
Otherwise, just target 1.18.

PR #1089, opened 2022-06-01
(1.16) Switch registration to use DeferredRegisters

lukebemish:
Similar to #1087, this switches registration to use deferred registers in the 1.16 branch. There are a few differences - notably, the BiomeLoadingEvent had to be used to add biome features - but for the most part it is quite similar.

*Sponsored by the [Modding Inquisition](https://github.com/TheModdingInquisition)*


maxanier:
Thank you for the 1.16 version.
I hope I will have time this weekend to review and merge both PRs

maxanier:
Looks good :)
I have very few minor changes. Can you enable "allow pushes from contributors" for both PRs?

Also, the formatting of several of the `core.Mod...` files (e.g. ModEntities) differs between 1.16 and 1.18. 
Long lines in one version are broken into two lines in the other version (mostly 1.16).
Generally, I prefer having long lines (that may be automatically wrapped by IntelliJ or so) instead of manually wrapped once.
Can you please unify this.

lukebemish:
Not quite sure how to allow changes for contributors on a PR; if this is some sort of repository setting, then I may not have the right permissions to do it on this repository (@Matyrobbrt, this might be something you can do). I'll try to unify those lines either tonight or tomorrow.

lukebemish:
Line wrapping should all be fixed, on both branches

Matyrobbrt:
@maxanier unfortunately, Github doesn't allow PRs whose head branches are from forks under an organization to enable that settings. I can work around that by giving you access to the repository.

maxanier:
Thank you for your contribution!
Will merge 1.18 later

PR #1094, opened 2022-06-18
[1.16] Update Altar of Inspiration model

lunofe:
| Old | New |
| - | - |
| image | image |

⚠️ Since MC 1.16 doesn't have candles yet, I've added the vanilla candle texture at `assets/vampirism/textures/block/candle.png`


maxanier:
java.io.FileNotFoundException: vampirism:textures/block/bloody_spruce_log.png

I think that texture was removed with the forest rework

maxanier:
Pls, also provide the voxelshapes/boundingbox code, if you are working with blockbench (see discord contributions pin)

lunofe:
Oh, forgot about the collision box thingy and didn't realise the missing texture as I was working with 1.8.8's assets - should be fixed now, I've switched the textures:

| Old | New (v2) |
| - | - |
| image | image |

However, I realised that there might be some minor stuff to edit as well, such as the material type:
https://github.com/TeamLapen/Vampirism/blob/900ebfce51a4dcf2c30b0336773f22256b2bc342/src/main/java/de/teamlapen/vampirism/blocks/AltarInspirationBlock.java#L57
The new model uses the cursed spruce log and a part of the dark spruce trapdoor, hence it should probably be changed to wood type as well.

---
And the crafting recipe should probably be changed also:
https://github.com/TeamLapen/Vampirism/blob/900ebfce51a4dcf2c30b0336773f22256b2bc342/src/main/java/de/teamlapen/vampirism/data/RecipesGenerator.java#L216
| Old | New |
| - | - |
| image | image<br>(Inspired by the vanilla barrel recipe) |

Perhaps you could even force the two new variants from the vampire forest as a type of wood, because you can definitely expect players to go there. *However, this could cause problems for players who use a different world generation and therefore might not have a vampire forest at all.*

Cheaterpaul:
I fixed a texture flickering of the candles and removed the cobwebs.
There was a texture glitch where the cobweb texture made the block see-through (could not find out why, strange behavior. You can add it back if you want).

~~The item model has a weird rotation, probably some missing rotations.~~


This will be merged for Vampirism 1.9, so https://github.com/TeamLapen/Vampirism/pull/1095 is not required.

PR #1095, opened 2022-06-18
[1.18] Update Altar of Inspiration model

lunofe:
| Old | New |
| - | - |
| image | image |


lunofe:
1.18 is not on par with 1.16 yet and lacks both https://github.com/TeamLapen/Vampirism/pull/1028 & https://github.com/TeamLapen/Vampirism/pull/1016, which contain textures this model rework is relying on. See https://github.com/TeamLapen/Vampirism/pull/1094 for main discussion.

maxanier:
As Paul mentioned, this will be done with 1.19 port. You might have to remind us to update the recipe to use the candles, though

PR #1120, opened 2022-08-12
Migrate vampirism crossbows to work like the vanilla crossbow

Cheaterpaul:
Because the vanilla crossbow handles everything with static methods, i adapted from them and modified the relevant parts.

Npcs (hinter/hunter minion) that hold a crossbow will always use it unless the target is in a sqrt(3) radius.

TODO:

- [x] deal with ammo consumption of npcs
  - [x] should minions consume arrows?
- [x] add small missing piece at the unloaded enhanced double crossbow model


Cheaterpaul:
The only missing part is if the minions should consume arrows or not. But that may need a more advanced AI to switch between melee and ranged depending on the arrows in their inventory.

Cheaterpaul:
minions can no longer use semi-auto crossbows unless the lord has the minion technology skill unlocked

PR #1174, opened 2023-01-23
Allow external modification of the minion visibility & capabilities

Cheaterpaul:
Currently the `PlayerBodyOverlayLayer` uses the visibility of the minion's `ModelPart`s to render the minion skin and player overlay separately. Bus this causes issues with the external modification of the `ModelPart`'s visibility.

With this change the vampire and hunter minions will have a dummy model that can be modified by 3rd parties but will not be rendered. The `PlayerBodyOverlayLayer` will then take over all rendering of the minion and the optional player overlay.

It will render the head separately from the body in case the player's model should be rendered. This will consider all changes to the visibility or rotations of the minion's dummy model.

Additionally the visibility of the hat will be considered when rendering the hat from the `HunterEquipmentLayer`

### DIFF
--- a/src/main/java/de/teamlapen/vampirism/entity/minion/MinionEntity.java
+++ b/src/main/java/de/teamlapen/vampirism/entity/minion/MinionEntity.java
@@ -62,9 +62,7 @@ import org.apache.logging.log4j.Logger;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.List;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
 import java.util.function.Predicate;
 
 public abstract class MinionEntity<T extends MinionData> extends VampirismEntity implements IPlayerOverlay, ISyncable, ForceLookEntityGoal.TaskOwner, de.teamlapen.vampirism.api.entity.minion.IMinionEntity, IEntityAdditionalSpawnData {
@@ -369,12 +367,12 @@ public abstract class MinionEntity<T extends MinionData> extends VampirismEntity
     public void onRemovedFromWorld() {
         if (playerMinionController != null) {
             playerMinionController.checkInMinion(this.minionId, this.token);
+            this.minionData.updateEntityCaps(this.serializeMinionCaps());
             this.minionData = null;
             this.playerMinionController = null;
         }
         super.onRemovedFromWorld();
     }
-
 
     @NotNull
     @Override
@@ -543,6 +541,7 @@ public abstract class MinionEntity<T extends MinionData> extends VampirismEntity
      * Can be called client and server side
      */
     protected void onMinionDataReceived(@NotNull T data) {
+        this.deserializeCaps(data.getEntityCaps());
     }
 
     @NotNull
@@ -598,7 +597,7 @@ public abstract class MinionEntity<T extends MinionData> extends VampirismEntity
         }
     }
 
-    private void handleLoadedMinionData(@NotNull T data) {
+    public final void handleLoadedMinionData(@NotNull T data) {
         this.getAttribute(Attributes.MAX_HEALTH).setBaseValue(data.getMaxHealth());
         super.setHealth(data.getHealth());
         super.setCustomName(data.getFormattedName());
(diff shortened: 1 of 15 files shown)


Cheaterpaul:
~~@maxanier regarding [99152e1](https://github.com/TeamLapen/Vampirism/pull/1174/commits/99152e1fea4aa9af330824326221212f4a1abe30)~~

~~I presume this was not intentionally left out. Does this have any side effects that we do not want to have? Of cause there can be some bad behavior, but it is more beneficial to allow mods to have the same capability for the same minion.~~

src/main/java/de/teamlapen/vampirism/entity/minion/MinionEntity.java
```diff
@@ -602,6 +602,7 @@ private void handleLoadedMinionData(@NotNull T data) {
         this.getAttribute(Attributes.MAX_HEALTH).setBaseValue(data.getMaxHealth());
         super.setHealth(data.getHealth());
         super.setCustomName(data.getFormattedName());
+        this.deserializeCaps(data.getEntityCaps());
```
Cheaterpaul: Saving an loading all available caps may introduce side effects that we do not want. E.g. if a cap kills the minion, loading the cap may constantly kill them again.

Cheaterpaul:
My solution now is to only save the `armourers_workshop:entity-skin-provider` capability.

src/main/java/de/teamlapen/vampirism/entity/minion/management/PlayerMinionController.java
```diff
@@ -208,6 +209,8 @@ public MinionEntity<?> createMinionEntityAtPlayer(int id, @NotNull Player p) {
                 } else {
                     m.claimMinionSlot(id, this);
                     m.copyPosition(p);
+                    // load from the saved data, but do not check it out in case the entity is not loaded
+                    ((MinionEntity<MinionData>)m).handleLoadedMinionData(minions[id].data);
```
Cheaterpaul: @maxanier could this be a problem?

If the entity is added to the world it is checked out (and loaded from the minion data) otherwise the entity is removed.

src/main/java/de/teamlapen/vampirism/entity/minion/MinionEntity.java
```diff
@@ -610,4 +609,27 @@ private void handleLoadedMinionData(@NotNull T data) {
...
+        } else {
+            return new CompoundTag();
+        }
+    }
+
+    /**
+     * @return all allowed capability identifiers
+     */
+    protected Collection<String> getAllowedCapTags() {
+        return Collections.singleton(new ResourceLocation("armourers_workshop", "entity-skin-provider").toString());
```
SAGESSE-CN, line 632: if you wish to use a whitelist, I recommend using the data pack design.

SAGESSE-CN: example:

in some mod, define a json like the tags:
resources/data/vampirism/minion.json 
```json
{
"capabillities": [
  "armourers_workshop:entity-skin-provider" 
 ]
}
```

Cheaterpaul: generally yes, but I currently do not see the need for it

src/main/java/de/teamlapen/vampirism/entity/minion/management/PlayerMinionController.java
maxanier: What is the benefit of this? It is loaded later anyway.
But if it is beneficial, this is probably fine. Maybe make sure the data is read only

maxanier:
what exactly is the problem with external modifications to the minion visibility at the moment. In theory the layer system should consider any visibility changes made to the model, shouldn't it?

maxanier:
If you think this is the best solution, feel free to merge it, I don't have time to look into this deeper

Cheaterpaul:
> what exactly is the problem with external modifications to the minion visibility at the moment. In theory the layer system should consider any visibility changes made to the model, shouldn't it?

Be cause we change the models visibility this does not work by default
https://github.com/TeamLapen/Vampirism/blob/f1caf610786d128f159f6055cebdf696b549d00f/src/main/java/de/teamlapen/vampirism/client/renderer/entity/layers/PlayerBodyOverlayLayer.java#L34-L42

PR #1179, opened 2023-02-15
Updated Repurposed Structures compat pieces

TelepathicGrunt:
Make their decor jigsaw block now properly point to the right pool file (I did a rename a long time ago to go from repurposed_structures:village/ to repurposed_structures:villages/ )

Also changed swamp piece to use mangrove blocks and fixed roof of badlands house to look better

image
image


Cheaterpaul:
Thanks again for this contribution and taking the time to change it.

PR #1181, opened 2023-02-18
Blood fog diffuser

### LINKED ISSUE #1104: [Feature]: Bloodfog Diffuser
### What?

Similarly to how Hunters get the Garlic Diffuser to protect themselves from Vampires, I propose that Vampires get a Bloodfog Diffuser that blankets the surrounding area in fog, allowing them to go about their business during the day within their territory as well as providing an aesthetic appeal. 

It would be crafted similarly to the Garlic Diffuser, except with a Diffuser Core that would be need to be made by combining Wool with a bottle of Pure Blood in a crafting grid. Purity 1-4 blood will create the Bloodfog Diffuser Core, while Purity 5 blood will create the Enhanced Bloodfog Diffuser Core, a more potent form of the core that affects a larger area. While within the area, the skybox and fog would take on a red tint similar to Vampire Forests, but less extreme. Right-clicking the diffuser with a plain Water Bottle would change the fog to a more natural tint similar to the Vampire villages, and right-clicking with a Blood Bottle would change it back, allowing the player to choose the aesthetic of their base.

### Mod experience

I believe this would affect mod balance positively, as it would enable a sort of soft "territory claim" system beyond just villages, in both a PvP sense and a roleplaying sense - vampires are classically territorial creatures, after all. It wouldn't give Vampire players any new abilities or buffs either, keeping in the spirit of the Vampire's power being contained within their skill tree and weapons, simply allowing them to roam about their own homes without fear of the sun. For PvP-enabled servers, it would also function as a tell for Hunter players, and perhaps NPC Hunter scouts could rarely spawn within chunks affected by a Bloodfog Diffuser similar to Pillager patrols.

Plus, the aesthetic of a vampire manor coated in foreboding fog is just really cool. :B
I know there's already the umbrellas mod, but there's no reason the two ideas can't co-exist, right? Temporary mobile protection vs permanent, fixed protection.

### Contributions

I have no contributions at present.

Cheaterpaul:
depends on #1278
close #1104

Adds a fog diffuser that emits fog for vampires covering 5x5 chunks.

The diffuser needs to be interacted with once with pure blood to start it up.

Todo:
- [x] model/textures
- [x] recipe
- [x] balancing


maxanier:
Nice!
Regarding recipe/balancing I have the following thought:
It requires some unique drop from the mother. Thereby it should be fairly rare and hard to obtain. Making it cover maybe 3x3 or maybe a "circle" with 5 chunks of diameter should cover a small base. Maybe it could even be larger.

I am unsure about the boosting part though. It seems very annoying to always remember to refill your generator. The fog is something you want to have long term, so you would have to constantly refill it.

supermike1999:
How rare would this mother/boss be? like a special boss that only spawns rarely, or one that you force spawn in like the wither/ender dragon? for the range of it, maybe have it a 5x5 area, but you can upgrade it to a better version of what a vampire controlled village totem gives.

I agree with the idea of you shouldn't need to consistently refill the diffusor, since garlic diffusors have the same idea of not needing to be refilled.

Cheaterpaul:
The diffuser now has a constant range but it needs to be started up by interacting with a pure blood. I did not want the diffuser to work out of the box, so it requires "fuel". But after inserting fuel the diffuer works forever.

This behaves a little like the beacon which needs to be charged with an item to work.

Cheaterpaul:
I went through the garlic diffuser code and noticed that garlic diffusers do not really work as they look.

- the startup progress bar shows should always charge up to 100% before the effect starts, but this does not apply depending on the context. When placing the diffuser the progress already starts at 100% which is set to 77% when fueling it.
- ~~The fueled state decreases over time, but nothing happens when it reaches 0%. Even after a world reload the garlic diffuser emits garlic when having 0% fuel.~~

This should be addressed by either change the ui or by correctly implement the process the user expects from the ui.

This should be addressed regardless of this PR

TheGridExpert:
Is it a thing for hunters?

PR #1195, opened 2023-04-03
add no blood effect when player has no food blood left

Cheaterpaul:
Vampires no longer can starve.
When they have no more blood they will receive several de-buffs on their attributes.

Currently modified attributes:
- movement speed x 0.8
- attack speed x 0.8
- sun damage x 1.2
- armor toughness x 0.7


Cheaterpaul:
@maxanier what do you think about these debuffs or do you have other ideas?

PR #1206, opened 2023-04-23
Leveling blocks feedback

Cheaterpaul:
The leveling block/entity menus now give feedback what items are missing, this makes use of the new `ItemCombinerMenu` class.

The game now prints out different messages when a player interacts with a faction block:
- wrong faction: "This seem unfamiliar"
- incomplete requirements: "You have not learned how to use this"

Additionally the vampire and hunter leveling configurations are rewritten


maxanier:
Looks nice.
While you are at it, maybe you could change the error message when there are not sufficient structure points, to something more meaningful. Like "missing x structure points" or "not enough pillars" or similar.

PR #1213, opened 2023-05-13
make vampire teleport into a dash like movement

Cheaterpaul:
## Teleport

Instead of teleporting the player instantly to the target, the player will now make up to 6 small teleports, which looks like a forward movement.

https://github.com/TeamLapen/Vampirism/assets/17816456/075b5039-ac2c-4963-8189-c750fbce4733

https://github.com/TeamLapen/Vampirism/assets/17816456/b8cc53db-48d2-405c-8158-91523792eaa3

## Additional functionality

This also intro a dispatched action system. Now we can dispatch action to the faction player which will be executed per tick.

## TODO

- [x] ensure correct teleportation when plying on a server


maxanier:
Looks awesome!
LastingAction was not suitable?

A few unrelated things I noticed while testing:
- At level 14 you don't even have enough skill points to unlock teleport
- Somehow I am not able to trigger the action by pressing `5`. Other actions with lower numbers work. Similar for the minion control. Only small numbers work

Cheaterpaul:
> LastingAction was not suitable?

We need to store some values in case the player disconnects mid-teleport. Actions do not support custom states. We would need to store the target (and some more) somewhere and I would rather store them context specific instead of adding another field somewhere.

> At level 14 you don't even have enough skill points to unlock teleport

you need to create a new world or change the skill points per level config option the the new default (`2`)

This might need to be addressed by modifying the config value when updating to Vampirism 1.10 

> Somehow I am not able to trigger the action by pressing 5

It seem not to be possible to select the skill with the highest number by pressing the number. This seem to be a problem with the skill screen

lusifer173:
It kinda looks weird to just quickly move in a direction and then a sudden puff of black smoke appears where you were. You should turn into a cloud of black smoke, then dash. It could play more into the idea that vampires can turn into mist.

maxanier:
Agreed. There should at least be a trail of smoke particles along the way

PR #1220, opened 2023-06-25
Vampirism web api

Cheaterpaul:
- swap contributor api
- allows to enable/disable features remotely in the api
- sends basic telemetry data, if it is not disabled in minecraft and vampirism
  - mod version
  - minecraft version
  - installed mods
  - server/client
- automatic update of config values and contributors if pushed to the default branch


src/main/java/de/teamlapen/vampirism/misc/SettingsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+    @Nullable
+    public String getSettingsValue(String key) {
+        return this.settingValues.get(key);
+    }
+
+    public @NotNull Optional<String> getSettingsValueOpt(String key) {
+        return Optional.ofNullable(this.settingValues.get(key));
+    }
+
+    public boolean isSettingTrue(String key) {
```
maxanier: Maybe it would be better to have a "default" parameter. Or just get rid of the nullable methods completely and only have optionals. (Optional<String> and Optional<Boolean> and Optional<Integer>)

src/main/java/de/teamlapen/vampirism/misc/SettingsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+    public CompletableFuture<Map<String, String>> getSettingValuesAsync(String modid) {
+        return get("config/list?modid=" + modid).thenApply(x -> GSON.fromJson(x, TypeToken.getParameterized(Map.class, String.class, String.class).getType()));
+    }
+
+    public CompletableFuture<Collection<Supporter>> GetSupportersAsync() {
+        Type type = TypeToken.getParameterized(List.class, Supporter.class).getType();
+        return get("supporter/list").thenApply(x -> GSON.fromJson(x, TypeToken.getParameterized(List.class, Supporter.class).getType()));
+    }
+
+    public CompletableFuture<Collection<Supporter>> GetSupportersAsync(String modid) {
```
maxanier: Upper case function name?

src/main/java/de/teamlapen/vampirism/misc/SettingsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+        return GetSupportersAsync().thenApply(x -> x.stream().filter(y -> y.faction().getNamespace().equals(modid)).collect(Collectors.toList()));
+    }
+
+    private CompletableFuture<String> get(String path) {
+        return get(this.apiVersion, path);
+    }
+
+    private CompletableFuture<String> get(String apiVersion, String path) {
```
maxanier: Why is api version separate from base url?
If we add a new api version on the server, the mod should completely switch to the new api, right?

src/main/java/de/teamlapen/vampirism/misc/SettingsProvider.java
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    private Map<String, String> checkSettings(Map<String, String> settings, Throwable error) {
+        if (error != null) {
+            LOGGER.error("Failed to retrieve settings from server", error);
+        }
+        if (VampirismMod.inDev || settings == null) {
+            try {
+                InputStream inputStream = VampirismMod.class.getResourceAsStream("/config_values.json");
```
maxanier: Maybe name file `default_remote_config`  or `fallback_api_values`. Also IOException text below is copy pasta

maxanier (commented):
Nice improvement

PR #1231, opened 2023-07-15
Update Forge version

Mgazul:
Fixed incompatibilities caused by changes in forge-47.1.6

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ org.gradle.daemon=false
 org.gradle.jvmargs=-Xmx3G
 #29.07.14 17:00
 minecraft_version=1.20.1
-forge_version=47.1.0
+forge_version=47.1.21
 mcp_mappings=1.19.3-2022.12.18-1.20.1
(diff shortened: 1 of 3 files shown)


Cheaterpaul:
Why the hell did they add a breaking change. They fixed it in 47.1.29 but only non functional.

Thanks for bringing this up and implementing this.

TelepathicGrunt:
@Cheaterpaul due to Neoforged being a thing and just about all of the staff that was under Lex moved to neo, Lex has now free reign over Forge. To show that no one was holding him back before, he has been merging in PRs without fully testing them or reviewing them. Picture that just about sums up why me and several modders are restricting our Forge version to 47.1.3 and older as these Forge versions are stable. Modpacks are probably going to build against 47.1.3 or the recommended build of 47.1.0 so might as well target this and bee free of anymore future breaking changes that Forge could get
Untitled

Cheaterpaul:
Regarding the changes see #1236

Issue #1235, opened 2023-07-19
[Suggestion]: Lower forge version to <47.1.3


Gronglegrowth:
### What?

**This is not a mod feature request**, but it is the closest template I could find.

**After Forge 47.1.3, there have been numerous breaking changes and bugs introduced in forge** (as a result of merging countless changes without bug-checking, _presumably_), and **many mods as a result have stuck to 47.1.0**, or .3 at the highest.

If one were to use the current version (1.9.1-beta2) of vampirism for 1.20 they would have to sacrifice use of many other mods, because it depends on 47.1.21 at minimum. **I am requesting you lower the forge version to 47.1.3**, at the highest.

If you are unfamiliar with what is going on with forge, which I assume so, [consider reading this](https://neoforged.net/news/theproject/). A tl;dr is, the forge devs (except for Lexmanos) forked into a new loader. As a result of this, forge versions past the aforementioned 47.1.3 are unreliable, as they are not maintained nor verified by the former devs, only Lexmanos and possibly whomever he has recruited since.

vectorwing:
Just to add on to this - yes, this does seem to be the case. Forge builds past 47.1.3 have introduced breaking changes due to a lack of proper triage, as the original team has left. TelepathicGrunt has replied under a PR with a picture of currently-detected breaking changes, so I figured I'd refer to it here as well:

[#1231](https://github.com/TeamLapen/Vampirism/pull/1231#issuecomment-1641186391)

Basically, the RB state of 47.1.x isn't very trustworthy anymore, so requiring a higher minimum version might cause unrelated bug reports, unless such versions fix a Vampirism bug from before it.

Cheaterpaul:
I am completely aware of the current situation and was holding a RB back in case there was more trouble on the way. But given the current state of Forge #1236.

In Vampirism, the `EntityEvent.EyeHeight`/`EntityEvent.Size` event is crucial and must work properly. Unfortunately, this feature is broken starting with Forge 46.1.6. As result, I had to upgrade to a newer Forge version in order to resolve this issue.

I agree that Vampirism should support Forge versions prior to 47.1.6, to allow it to be used alongside other mods that require a prior version or users that do not wish to use newer versions. To achieve this compatibility, Vampirism may need to either remove the feature, specify a version range of [46.1.0,46.1.3], or utilize reflection to support both newer and older forge versions.

Cheaterpaul:
The newest version only requires 47.1.0

PR #1236, opened 2023-07-19
Make Vampirism compatible with newer and older Forge versions

Cheaterpaul:
As reasoning https://github.com/TeamLapen/Vampirism/pull/1231#issuecomment-1641186391 and https://github.com/TeamLapen/Vampirism/issues/1235

We have multiple choices here:
1. use Forge > 46.1.29
    - will work
    - forces the user to use newer forge versions
    - forces Vampirism to be removed / not added to modpacks
2. restrict Vampirism to Forge [46.1.0, 46.1.3]
    - will work
    - user will complain why Vampirism does not support newer version
    - limits the users freedom
3. try to do both
   - all forge version will be supported
   - the users can choose what to do
   - if someone complains / reports a bug we can always tell them to use Forge 46.1.3

While I do not like any of these choices I think the 3rd one is the best one given the circumstances.

This solution is based on Forge 46.1.0, keeps the old methods (we do not care about Forge `[46.1.6-46.1.29)`) and uses reflections to register the eye height event for newer versions

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -5,7 +5,7 @@ org.gradle.jvmargs=-Xmx3G
 # forge
 minecraft_version=1.20.1
 minecraft_version_range=[1.20.1,1.21)
-forge_version=47.1.21
+forge_version=47.1.0
 forge_version_range=[47,)
 loader_version_range=[47,)
 mappings_channel=parchment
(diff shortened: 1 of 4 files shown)


maxanier:
I would always prefer to stick with recommended build. Anyone using most recent MinecraftForge can still use .beta.2.
Once the dust has settled we will migrate to NeoForge, so eventually this won't be an issue anymore.
I don't think we should actively support latest MinecraftForge

maxanier:
We should restrict the Forge version range then and maybe manually change the CurseForge Displaytitle to make things clear

Issue #770, opened 2020-09-09
[Feature] Vampires should be treated as undead


ryleu:
## What would you like to have changed or added to Vampirism
Vampires should work like other undead mobs in regards to instant damage / health. Zombies and skeletons get hurt by healing and healed with harming, so vampires should too.

## Does this make the mod experience more balanced?
It doesn't mess with the balance very much, as hunters can use healing and vampires can use harming.

maxanier:
This was requested many times and I agree that it would be quite cool.
However, after putting some more thought into it, this isn't viable from a modding point of view.
Inverting these effects could very easily mess up other mods that rely on instant damage or poison damaging a player.
I don't want to risk that.

Since it isn't viable for players, it won't be added for NPCs either

ryleu:
That makes a lot of sense. The only way you could realistically make this work is with some sort of additional logic around potions and splash potions that would end up being either really expensive or really hard to implement.

Thanks for giving my idea a read and have a nice day!

MarioSMB:
Pardon the bump, but I'm curious - is this not a simple case of extending the player to have the UNDEAD MobType when they're a vampire? Any mods that don't support it should probably be notified that they're doing some improper checks.

I'm not entirely sure how the Forge and Minecraft coding ecosystem works so I may be way off, just thought I'd mention it!

ryleu:
you might be able to mixin into the check for MobType and have it also return UNDEAD if an entity is a vampire, but that sounds like it would be an expensive check because of how often it'd run

maxanier:
Yes, it should theoretically be possible via Mixins. Performance would not be great, but I guess not too bad either.
Still I don't think it is worth the trouble. 
Could try implementing this as an optional feature, so people who really want this, can try if it works.  However, from past experience (and the lack of feedback) little users seem to use any of Vampirism's optional features

MarioSMB:
I am genuinely surprised Forge offers no API hooks for mob types, but I do believe it would be worth trying - some things like drowning protection (and the desired potion behaviour) would be simplified as the game itself provides them, and it would make weapons with the smiting enchantment effective against vampires out of the box. 
Perhaps it's a matter of weighing the benefits vs the costs; if mixins are really slow then indeed it might not offset the benefits of allowing undead checks to work.

Regarding optional features; for what it's worth I don't even know what optional features the mod provides! It may help to announce or list them somewhere easy to find - perhaps a dedicated section on the wiki.  
Food (or blood in this case) for thought!

PR #1250, opened 2023-08-05
Add mixin for MobType

maxanier:
@Cheaterpaul What do you think about this approach?
Should not really impact performance, right?
Instead of instance of check, there is a delegate method and overwrite.

My idea is to introduce #770 as optional in 1.20/1.9 and let e.g. Mario test it.
In the future we could consider making it default.

### DIFF
--- a/src/api/java/de/teamlapen/vampirism/api/VReference.java
+++ b/src/api/java/de/teamlapen/vampirism/api/VReference.java
@@ -43,8 +43,9 @@ public class VReference {
      */
     public static MobCategory VAMPIRE_CREATURE_TYPE = MobCategory.create("vampirism_vampire", "vampirism_vampire", 30, false, false, 128);
     /**
-     * Vampire creatures have this creature attribute.
+     * Vampire creatures have this creature attribute. Note: There is a config option that makes Vampirism use UNDEAD type instead
      * Don't know why this exists alongside EnumCreatureType, but this is used by enchantments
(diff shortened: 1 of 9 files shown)


Cheaterpaul:
I see no problem in the implementation. This should have no performance impact, only when called on the player and there we already have a VampirismAttributes class to minimize it.

But the main problem with making vampires undead is that all other game features will treat the vampire as undead. Yes this is the intention of this change, but before doing this we should make sure that no vanilla feature or vampirism feature have side effects with this change.

The smite enchantment for example deals 15.5 extra damage to undead entities which is simply to much against a player. I am not aware of any other side effects, but I suspect there are more.

supermike1999:
Comparing it, you need Sharpness level 24 to match Smite 5 when attacking a undead mob. I simulated what it looks like in combat. (Normal player with [Smite 5 Netherite word] vs Maxed level Vampire with natural armor)

https://youtu.be/6kL4d2Wa6Jg (Video was way over 10mb limit)

maxanier:
Yeah, it will probably be a bit unfair, but people are asking for it, so why not make it an (off by default) option.
Maybe we can consider mitigating the balancing issues somehow later.

I will extend this configuration to vampire NPCs and merge eventually.

Cheaterpaul:
@Cheaterpaul add as experimental feature with enough warnings and take a look where this would be a problem.

Smite is a problem while harming / healing potions should be fine.

PR #1295, opened 2023-12-03
Events for Vampire Blood Drinking

TheDrOfDoctoring:
This adds three new events, `PlayerDrinkBloodEvent`, `EntityDrinkBloodEvent` and `BiteFeedEvent`. 

`PlayerDrinkBloodEvent` is fired whenever a Vampire Player drinks blood from any source which includes blood bottles, mob biting and vampire food items such as human hearts. The amount of blood gained, saturation and whether the remaining blood is used or not can be changed when the event is caught.

`EntityDrinkBloodEvent` is fired whenever a non-player Vampire Entity drinks blood from any source, although this will always be from mob biting under regular circumstance.

`BiteFeedEvent` is fired whenever a Vampire Player drinks blood from a mob, which exposes the entity being drained from in the event. Only the amount of blood gained and saturation can be changed here when the event is caught, the `PlayerDrinkBloodEvent` will be fired after this as `drinkBlood` is called.

The potential use case of these events are to decrease the amount of blood a vampire may get under some conditions, or to check how much blood a vampire is draining. (This would be the use case in the Vampiric Ageing addon, which currently uses a mixin to handle this). 

The final commit includes some slightly larger changes as it includes an enum for blood sources which contains 5 values:
```
    BITE_FEED,
    VAMPIRE_FOOD,
    BLOOD_BOTTLE,
    CONTAINER,
    OTHER
```

`BITE_FEED` refers to any source that came from drinking another mob's blood.
`VAMPIRE_FOOD` refers to any source that came from eating vampire food such as human hearts.
`BLOOD_BOTTLE` refers to drinking from a blood bottle.
`CONTAINER` refers to blood drunk from containers.
`OTHER` refers to blood drunk from any source not covered by the previous ones, such as the blood given by the level up ritual.

These enum values are then available through the event, allowing to check for specific cases. For example, under some circumstances a modder may want a vampire to gain less blood from human hearts and have them gain more from directly feeding from mobs. The use case for this is mainly for something I am planning to do with a future addon, although I imagine it could be useful for other purposes.


Cheaterpaul:
I removed the EnumBloodSource in favor of a context object. Additionally i did some QoL changes to the code

PR #1296, opened 2023-12-06
Events for Actions

TheDrOfDoctoring:
This adds 4 new events to the action system to allow for some degree of modification of existing actions. These actions are `ActionActivatedEvent`,  `ActionDeactivatedEvent`, `ActionUpdateEvent` and `PreActionActivatedEvent`.

`PreActionActivatedEvent` allows an action to be cancelled before it is activated.

`ActionActivatedEvent` allows the cooldown of an action to be changed, as well as the duration for lasting actions.

`ActionDeactivatedEvent` allows for the cooldown of an action to be changed before it is deactivated.

`ActionUpdateEvent` allows is fired for every action update, but does not allow for any part of the action to be changed.

Additionally, lasting actions that add potion effects (such as sunscreen or rage) have been moved to applying the effect every 20 ticks under `onUpdate`. Previously, the potion effects were only added once when first activated. The regeneration vampire action also has been moved to a lasting action due to all the other similar actions being lasting actions already.

### DIFF
--- a/src/api/java/de/teamlapen/vampirism/api/event/ActionEvent.java
+++ b/src/api/java/de/teamlapen/vampirism/api/event/ActionEvent.java
@@ -1,0 +1,129 @@
+package de.teamlapen.vampirism.api.event;
+
+import de.teamlapen.vampirism.api.entity.player.IFactionPlayer;
+import de.teamlapen.vampirism.api.entity.player.actions.IAction;
+import net.minecraftforge.eventbus.api.Cancelable;
+import net.minecraftforge.eventbus.api.Event;
+import org.jetbrains.annotations.NotNull;
+
+@SuppressWarnings("unused")
+public abstract class ActionEvent extends Event {
+
+    @NotNull
+    private final IFactionPlayer<?> factionPlayer;
+    @NotNull
+    private final IAction<?> action;
+    public ActionEvent(@NotNull IFactionPlayer<?> factionPlayer, @NotNull IAction<?> action) {
+        this.factionPlayer = factionPlayer;
+        this.action = action;
+    }
+    /**
+     * @return The FactionPlayer who activated the action.
+     */
+    public @NotNull IFactionPlayer<?> getFactionPlayer() {
+        return this.factionPlayer;
+    }
+    /**
+     * @return The action the event is firing for.
+     */
+    public @NotNull IAction<?> getAction() {
+        return this.action;
+    }
+
+    /**
+     * Posted before an action fires. Use this to modify the cooldown or duration of an event
+     */
+    public static class ActionActivatedEvent extends ActionEvent {
+
+        private int cooldown;
+        private int duration;
+
+        public ActionActivatedEvent(@NotNull IFactionPlayer<?> factionPlayer, @NotNull IAction<?> action, int cooldown, int duration) {
+            super(factionPlayer, action);
+            this.cooldown = cooldown;
+            this.duration = duration;
+        }
+        /**
+         * @return The original cooldown of the action
+         */
+        public int getCooldown() {
+            return cooldown;
+        }
+        /**
+         * @param cooldown the new cooldown of the action.
+         */
+        public void setCooldown(int cooldown) {
+            this.cooldown = cooldown;
+        }
+        /**
+         * @return The original cooldown of the duration, this will return 0 if the action does not implement ILastingAction.
+         */
+        public int getDuration() {
+            return duration;
+        }
+        /**
+         * @param duration the new duration of the action.
+         */
+        public void setDuration(int duration) {
+            this.duration = duration;
+        }
+    }
+    /**
+     * Posted when an action deactivates, either when deactivated manually or when out of time. As regular actions instantly deactivate, this only fires for actions that implement ILastingAction.
+     */
+    public static class ActionDeactivatedEvent extends ActionEvent {
+        private final int remainingDuration;
+        public ActionDeactivatedEvent(@NotNull IFactionPlayer<?> factionPlayer, @NotNull IAction<?> action, int remainingDuration) {
+            super(factionPlayer, action);
+            this.remainingDuration = remainingDuration;
+        }
+        /**
+         * @return The remaining cooldown of the action.
+         */
+        public int getRemainingDuration() {
+            return remainingDuration;
+        }
+
+    }
+    /**
+     * Posted when an action deactivates, either when deactivated manually or when out of time. As regular actions instantly deactivate, this only fires for actions that implement ILastingAction.
+     */
+    public static class ActionUpdateEvent extends ActionEvent {
+        private final int remainingDuration;
+        private boolean shouldDeactivate;
+        public ActionUpdateEvent(@NotNull IFactionPlayer<?> factionPlayer, @NotNull IAction<?> action, int remainingDuration, boolean shouldDeactivate) {
+            super(factionPlayer, action);
+            this.remainingDuration = remainingDuration;
+            this.shouldDeactivate = shouldDeactivate;
+        }
+        /**
+         * @return The remaining cooldown of the action.
+         */
+        public int getRemainingDuration() {
+            return this.remainingDuration;
+        }
+        /**
+         * @return Whether the lasting action should be deactivated
+         */
+        public boolean shouldDeactivate() {
+            return this.shouldDeactivate;
+        }
+        /**
+         * @param deactivate If set to true, the action will deactivate. If set to false, the action will not deactivate at all on that update.
+         */
+        public void setDeactivation(boolean deactivate) {
+            this.shouldDeactivate = deactivate;
+        }
+
+    }
+    /**
+     * Posted before a player activates an action, can be used to prevent an action being fired.
+     */
+    @Cancelable
+    public static class PreActionActivatedEvent extends ActionEvent {
+        public PreActionActivatedEvent(@NotNull IFactionPlayer<?> factionPlayer, @NotNull IAction<?> action) {
+            super(factionPlayer, action);
+        }
+    }
+
+}
--- a/src/api/java/de/teamlapen/vampirism/api/event/ActionEvent.java
+++ b/src/api/java/de/teamlapen/vampirism/api/event/ActionEvent.java
@@ -31,7 +31,7 @@ public abstract class ActionEvent extends Event {
     }
 
     /**
-     * Posted before an action fires. Use this to modify the cooldown or duration of an event
+     * Posted after an action fires. Use this to modify the cooldown or duration of an event
      */
     public static class ActionActivatedEvent extends ActionEvent {
 
@@ -44,25 +44,25 @@ public abstract class ActionEvent extends Event {
             this.duration = duration;
         }
         /**
-         * @return The original cooldown of the action
+         * @return The original cooldown of the action, in ticks
          */
         public int getCooldown() {
             return cooldown;
         }
         /**
-         * @param cooldown the new cooldown of the action.
+         * @param cooldown the new cooldown of the action, in ticks.
          */
         public void setCooldown(int cooldown) {
             this.cooldown = cooldown;
         }
         /**
-         * @return The original cooldown of the duration, this will return 0 if the action does not implement ILastingAction.
+         * @return The original duration of the action, this will return 0 if the action does not implement ILastingAction.
          */
         public int getDuration() {
             return duration;
         }
         /**
-         * @param duration the new duration of the action.
+         * @param duration the new duration of the action, in ticks.
          */
         public void setDuration(int duration) {
             this.duration = duration;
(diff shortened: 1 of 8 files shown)


src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -264,13 +284,24 @@ public void saveToNbt(@NotNull CompoundTag nbt) {
...
+     * @param action  Action being toggled
+     * @param context Context holding Block/Entity the player was looking at when activating if any
+     *
+     */
     @Override
     public IAction.PERM toggleAction(@NotNull IAction<T> action, IAction.ActivationContext context) {
         ResourceLocation id = RegUtil.id(action);
         if (activeTimers.containsKey(id)) {
             deactivateAction((ILastingAction<T>) action);
+            this.activeTimers.removeInt(id);
```
Cheaterpaul, line 304: Why not move this in the `deactiveAction` method?

Cheaterpaul: Is there a reason why the remove is not in the `deactivateAction` method?

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -238,24 +260,22 @@ public void resetTimers() {
         for (ResourceLocation id : activeTimers.keySet()) {
             @SuppressWarnings("unchecked")
             ILastingAction<T> action = (ILastingAction<T>) RegUtil.getAction(id);
-            assert action != null;
-            action.onDeactivated(player);
+            deactivateAction(action, true);
         }
         activeTimers.clear();
         cooldownTimers.clear();
+        modifiedCooldownTimers.clear();
```
Cheaterpaul: the `modifiedDurationTimer` clear is missing

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -45,6 +47,16 @@ public class ActionHandler<T extends IFactionPlayer<T>> implements IActionHandle
...
+    /**
+     * Stores the modified cooldown of an action after it has been changed by an event. this is used to check the action cooldown instead of {@link de.teamlapen.vampirism.api.entity.player.actions.IAction#getCooldown(IFactionPlayer)}
+     * The values stored here are only changed when the modified cooldown is added, it is not decremented like the map for cooldown timers.
+     */
+    private final @NotNull Object2IntMap<ResourceLocation> modifiedCooldownTimers;
+    /**
+     * Stores the modified duration of an action, this is used to check the action duration instead of {@link de.teamlapen.vampirism.api.entity.player.actions.ILastingAction#getDuration(IFactionPlayer)} (IFactionPlayer)}
+     * The values stored here are only changed when the modified duration is added, it is not decremented like the map for duration timers.
+     */
+    private final @NotNull Object2IntMap<ResourceLocation> modifiedDurationTimer;
```
Cheaterpaul: I would move away from calling this modified, and fix the java doc accordingly. (Maybe `expectedCooldownTimes` or `expectedDurations`)

The cooldown / duration is stored here even if they are not modified by the event (which was my idea as well).

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -238,24 +260,22 @@ public void resetTimers() {
...
-            activeTimers.removeInt(RegUtil.id(action));
-        }
+        ILastingAction<T> lastingAction = (ILastingAction<T>) action;
+        deactivateAction(lastingAction, true);
         cooldownTimers.removeInt(RegUtil.id(action));
-        dirty = true;
-    }
+        modifiedCooldownTimers.removeInt(RegUtil.id(action));
 
+    }
```
Cheaterpaul: You can not assume that `action` is a lasting action here. The interface contract `void resetTimer(@NotNull IAction<T> action)` allows any action to be passed as `action`.

`dirty = true` is missing and the `modifiedDurationTimer#remove` call as well.

src/api/java/de/teamlapen/vampirism/api/event/ActionEvent.java
```diff
@@ -0,0 +1,144 @@
...
+        public void setDeactivation(boolean deactivate) {
+            this.shouldDeactivate = deactivate;
+        }
+
+    }
+    /**
+     * Posted before a player activates an action, can be used to prevent an action being fired.
+     */
+    @Cancelable
+    public static class PreActionActivatedEvent extends ActionEvent {
```
Cheaterpaul: Please remove the Pre event and merge it with the `ActionActivatedEvent` and rename it to `ActionActivateEvent`

src/api/java/de/teamlapen/vampirism/api/event/ActionEvent.java
```diff
@@ -0,0 +1,144 @@
...
+     * @return The action the event is firing for.
+     */
+    public @NotNull IAction<?> getAction() {
+        return this.action;
+    }
+
+    /**
+     * Posted after an action fires. Use this to modify the cooldown or duration of an event
+     */
+    public static class ActionActivatedEvent extends ActionEvent {
```
Cheaterpaul, line 40: make it cancel-able see below

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -107,11 +118,12 @@ public void extendActionTimer(@NotNull ILastingAction<T> action, int duration) {
...
-        if (activeTimers.containsKey(RegUtil.id(action))) {
-            return activeTimers.getInt(RegUtil.id(action)) / ((float) ((ILastingAction<T>) action).getDuration(player));
+        ResourceLocation id = RegUtil.id(action);
+        if (activeTimers.containsKey(id)) {
+            return activeTimers.getInt(id) / ((float) modifiedDurationTimer.getOrDefault(id, ((ILastingAction<T>) action).getDuration(player)));
         }
-        if (cooldownTimers.containsKey(RegUtil.id(action))) {
-            return -cooldownTimers.getInt(RegUtil.id(action)) / (float) action.getCooldown(player);
+        if (cooldownTimers.containsKey(id)) {
+            return -cooldownTimers.getInt(id) / (float) modifiedCooldownTimers.getOrDefault(id, action.getCooldown(player));
```
Cheaterpaul: We expect here that modifiedDuration/Cooldown is filled with the value. There is no need to provide a default value

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -279,16 +310,32 @@ public IAction.PERM toggleAction(@NotNull IAction<T> action, IAction.ActivationC
...
             IAction.PERM r = action.canUse(player);
             if (r == IAction.PERM.ALLOWED) {
+
                 if (action.onActivated(player, context)) {
                     ModStats.updateActionUsed(player.getRepresentingPlayer(), action);
+                    int duration = 0;
+                    if (action instanceof ILastingAction) {
+                        duration = ((ILastingAction<T>) action).getDuration(player);
+                    }
+                    ActionEvent.ActionActivatedEvent activationEvent = VampirismEventFactory.fireActionActivatedEvent(player, action, action.getCooldown(player), duration);
```
Cheaterpaul: move this event directly before the `action.onActivated` but store the duration and cooldown only after the activation.

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -279,16 +310,32 @@ public IAction.PERM toggleAction(@NotNull IAction<T> action, IAction.ActivationC
...
+                        modifiedDurationTimer.put(id, activationEvent.getDuration());
+                        duration = activationEvent.getDuration();
+                        activeTimers.put(id, duration);
                     } else {
-                        cooldownTimers.put(id, action.getCooldown(player));
+                        cooldownTimers.put(id, cooldown);
                     }
                     dirty = true;
+                } else {
+                    modifiedDurationTimer.removeInt(id);
```
Cheaterpaul: You do not need this. The entry should not exist

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -348,12 +417,17 @@ public boolean updateActions() {
             ILastingAction<T> action = (ILastingAction<T>) RegUtil.getAction(entry.getKey());
             assert action != null;
             if (newtimer == 0) {
-                action.onDeactivated(player);
-                cooldownTimers.put(entry.getKey(), action.getCooldown(player));
+                deactivateAction(action, true);
+                if(!cooldownTimers.containsKey(entry.getKey())) {
+                    cooldownTimers.put(entry.getKey(), modifiedCooldownTimers.getInt(RegUtil.id(action)));
```
Cheaterpaul: Just override the timer if it exists. 

use `entry.getKey()` instead of `RegUtil.id(action)`

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
```diff
@@ -348,12 +417,17 @@ public boolean updateActions() {
...
+                }
                 it.remove();//Do not access entry after this
+
                 dirty = true;
             } else {
-                if (action.onUpdate(player)) {
+                boolean shouldDeactivate= action.onUpdate(player);
+                boolean shouldReallyDeactivate = VampirismEventFactory.fireActionUpdateEvent(player, action, newtimer, shouldDeactivate);
+                if (shouldReallyDeactivate) {
```
Cheaterpaul: First of all make this a one liner.

Secondly the other question here is if we only want to allow mods to change the deactivation or if we want to give more control to the mods. E.g allow the the event handler to skip the `action.onUpdate(player)` call.

I could imagine a case where someone does not want the `action#onUpdate` to be called, but does not want to disable the action. This could be achieved by using the event's result.

DENY -> skip the `action#onUpdate` and deactivate the action if the event is set to do it.
DEFAULT -> ignore the event properties, call `action#onUpdate` and deactivate accordingly.
ALLOW -> call `action#onUpdate` but ignore the return value.

What are your thought about this. With this approach, however, the event handler has no access to the `action#onUpdate` return value unless they call it themselves and set the result to DENY.

src/api/java/de/teamlapen/vampirism/api/event/ActionEvent.java
```diff
@@ -0,0 +1,144 @@
...
+
+        private int cooldown;
+        private int duration;
+
+        public ActionActivatedEvent(@NotNull IFactionPlayer<?> factionPlayer, @NotNull IAction<?> action, int cooldown, int duration) {
```
Cheaterpaul, line 45: given that this event is used with no `ILastingAction` make the `#getDuration` either return -1 or change the return value into an OptionalInt, while throwing an exception when setting the duration it is is not a `ILastingAction`. 

0 may be confusing as it is technically a valid duration.

Cheaterpaul (changes requested):
Very nice 🎉🎉. That's about what i thought for such action events

I haven't tested it yet, but there where some minor issues with the implementation and some regarding the design. A lot of requested changed but mainly the following:

1. PreActionActivatedEvent
   To me there is no point in having two event that are called for activating an action if they can be rearranged and merged into one.
2. UpdateEvent
   I have no problem with this if it is only about deactivating the action prematurely. But it might be worth to invest more time on the event to make it more powerful.

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
TheDrOfDoctoring: Changed, although I haven't added`modifiedDurationTimer#remove` as it is called in `deactivateAction`

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
TheDrOfDoctoring: Yes, removing it in `deactivateAction` causes a crash `onUpdate` when it tries to iterate through `activeTimers`, however it doesn't cause any issues when removed here.

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
TheDrOfDoctoring: If I'm understanding correctly, overriding it here would cause the incorrect timer to be applied. For example, when the action is deactivated manually it needs to have a reduced cooldown based on the remaining duration. This wouldn't work here and would cause it to use the full cooldown. This is because the timer can't be directly removed in `deactivateAction`.

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
TheDrOfDoctoring: I agree with that, probably the better approach. I'm not sure if there's that much of a use case where the return value of `onUpdate` needs to be known. If it returns true, it will be deactivated anyway which could be checked with the deactivation event, although I suppose it may be useful to know if the action is going to continue. So, changing to your approach is probably best.

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
Cheaterpaul: I changed the `onUpdate` method

src/main/java/de/teamlapen/vampirism/entity/player/actions/ActionHandler.java
Cheaterpaul: I added a deactivate parameter to the update event

Cheaterpaul (approved):

Cheaterpaul:
🎉🎉

PR #1297, opened 2023-12-10
More Vampire Attributes

TheDrOfDoctoring:
This adds 3 new attributes for easier modification of certain vampire abilities that would previously require mixins to change. The attributes are `NEONATAL_FORM`, `DBNO_TIMER`, and `BLOOD_EXHAUSTION_INCLUDE_HEALING`.

`NEONATAL_FORM` attribute can be used to change the duration of the neonatal effect. 
`DBNO_TIMER` attribute can be used to change the resurrection timer.
`BLOOD_EXHAUSTION_INCLUDE_HEALING` is similar to the Blood Exhaustion attribute except it applies to exhaustion added for healing as well.

In order to avoid changing the default config value for the reduction skills (since in existing configs when updated the values wont be changed), the values in the config for `DBNO_TIMER` and `NEONATAL_FORM` are reduced by 1 when they are added via the skills.

Additionally, I'm not sure about having the two different blood exhaustion attributes. The main reason for it is that addons may want to significantly increase or decrease attribute, but this means that the exhaustion for healing doesn't change at all as reducing the exhaustion for healing by the same value can be a bit too much. 

However, as barely any exhaustion is drained except for healing a large change in the blood exhaustion attribute means exhaustion is essentially only changed when healing, causing the player to mostly stay at a full blood bar but with low saturation.
 
To a player, this can be unclear as the saturation is hidden and the only visible indicator they have is how full their blood bar is, which can be confusing as if the saturation is low they will only heal very slowly.

By default, the attribute makes no changes as it's set to 1, it just helps to give a bit more control over how blood is handled for addons, however I'm not sure if it might be better to just leave that to mixins.


src/main/java/de/teamlapen/vampirism/core/ModAttributes.java
```diff
@@ -22,6 +22,21 @@ public class ModAttributes {
      * Registered for EntityPlayer
      */
     public static final RegistryObject<RangedAttribute> BLOOD_EXHAUSTION = ATTRIBUTES.register("blood_exhaustion", () -> (RangedAttribute) new RangedAttribute("vampirism.blood_exhaustion", 1.0, 0.0, 10).setSyncable(true));
+    /**
+     * Allows modifying the duration of the neonatal effect.
+     * Registered for EntityPlayer
+     */
+    public static final RegistryObject<RangedAttribute> NEONATAL_FORM = ATTRIBUTES.register("neonatal_form", () -> (RangedAttribute) new RangedAttribute("vampirism.neonatal_form", 1.0, 0.0, Integer.MAX_VALUE).setSyncable(true));
```
Cheaterpaul: Can you rename the attribute to "neonatal_duration"

src/main/java/de/teamlapen/vampirism/core/ModAttributes.java
```diff
@@ -22,6 +22,21 @@ public class ModAttributes {
...
+    /**
+     * Allows modifying the length of the resurrection timer.
+     * Registered for EntityPlayer
+     */
+    public static final RegistryObject<RangedAttribute> DBNO_TIMER = ATTRIBUTES.register("dbno_timer", () -> (RangedAttribute) new RangedAttribute("vampirism.dbno_timer", 1.0, 0.0, Integer.MAX_VALUE).setSyncable(true));
```
Cheaterpaul: Can you rename the attribute to "dbno_duration"

src/main/java/de/teamlapen/vampirism/entity/player/vampire/VampirePlayer.java
```diff
@@ -441,11 +440,7 @@ public void increaseRemainingBarkTicks(int additionalTicks) {
     }
 
     public int getDbnoDuration() {
-        int duration = VampirismConfig.BALANCE.vpDbnoDuration.get() * 20;
-        if (this.skillHandler.isSkillEnabled(VampireSkills.DBNO_DURATION.get())) {
-            duration = Math.max(1, (int) (duration * VampirismConfig.BALANCE.vsDbnoReduction.get()));
-        }
-        return duration;
+        return Math.max(1, (int) player.getAttribute(ModAttributes.DBNO_TIMER.get()).getValue());
```
Cheaterpaul: just return `(int) player.getAttributeValue(ModAttributes.DBNO_TIMER.get()`

src/main/java/de/teamlapen/vampirism/entity/player/vampire/VampirePlayer.java
```diff
@@ -1053,10 +1048,7 @@ public void tryResurrect() {
             this.player.setForcedPose(null);
             this.player.refreshDimensions();
             this.sync(true);
-            int duration = VampirismConfig.BALANCE.vpNeonatalDuration.get() * 20;
-            if (this.skillHandler.isSkillEnabled(VampireSkills.NEONATAL_DECREASE.get())) {
-                duration = Math.max(1, (int) (duration * VampirismConfig.BALANCE.vsNeonatalReduction.get()));
-            }
+            int duration = Math.max((int) player.getAttribute(ModAttributes.NEONATAL_FORM.get()).getValue(), 1);
```
Cheaterpaul: same here

src/main/java/de/teamlapen/vampirism/config/BalanceConfig.java
```diff
@@ -330,8 +330,8 @@ public class BalanceConfig {
         vsBloodVisionDistanceSq = builder.comment("Squared blood vision distance").defineInRange("bloodVisionDistanceSq", 1600, 5, Integer.MAX_VALUE);
         vsSmallAttackDamageModifier = builder.comment("Damage added to base damage").defineInRange("smallAttackDamageModifier", 1d, 0, 10d);
         vsSmallAttackSpeedModifier = builder.comment("Basic skill - Weapon cooldown = 1/(oldvalue*(1+modifier))").defineInRange("smallAttackSpeedModifier", 0.15, 0, 3);
-        vsNeonatalReduction = builder.comment("Reduced percentage of the neonatal effect").defineInRange("neonatalReduction", 0.5, 0, 1);
-        vsDbnoReduction = builder.comment("Reduced percentage of the downed timer required to resurrect").defineInRange("dbnoReduction", 0.5, 0, 1);
+        vsNeonatalReduction = builder.comment("Reduced percentage of the neonatal effect").defineInRange("neonatalReduction", 0.5, -1024, 1024);
+        vsDbnoReduction = builder.comment("Reduced percentage of the downed timer required to resurrect").defineInRange("dbnoReduction", 0.5, -1024, 1024);
```
Cheaterpaul: What is the idea here?

The config values are applied to the attribute with `MULTIPLY_TOTAL`. It does not make sense to use values below `0` for this option

src/main/java/de/teamlapen/vampirism/entity/player/vampire/skills/VampireSkills.java
```diff
@@ -73,8 +73,8 @@ protected void getActions(@NotNull Collection<IAction<IVampirePlayer>> list) {
...
-    public static final RegistryObject<ISkill<IVampirePlayer>> NEONATAL_DECREASE = SKILLS.register("neonatal_decrease", () -> new VampirismSkill.SimpleVampireSkill(2, true));
-    public static final RegistryObject<ISkill<IVampirePlayer>> DBNO_DURATION = SKILLS.register("dbno_duration", () -> new VampirismSkill.SimpleVampireSkill(2, true));
+    public static final RegistryObject<ISkill<IVampirePlayer>> NEONATAL_DECREASE = SKILLS.register("neonatal_decrease", () -> new VampirismSkill.SimpleVampireSkill(2, true).registerAttributeModifier(ModAttributes.NEONATAL_FORM.get(), "74ecad40-5674-4ee0-8224-f871cdd6543d", () -> VampirismConfig.BALANCE.vsNeonatalReduction.get() - 1, AttributeModifier.Operation.MULTIPLY_TOTAL));
+    public static final RegistryObject<ISkill<IVampirePlayer>> DBNO_DURATION = SKILLS.register("dbno_duration", () -> new VampirismSkill.SimpleVampireSkill(2, true).registerAttributeModifier(ModAttributes.DBNO_TIMER.get(), "7e5b2fb1-ba74-4838-83e1-68fd419af787", () -> VampirismConfig.BALANCE.vsNeonatalReduction.get() - 1, AttributeModifier.Operation.MULTIPLY_TOTAL));
```
Cheaterpaul: This uses the `vsNeonatalReduction` config option

Cheaterpaul (changes requested):
I have not finished reviewing all changes especially the exhaustion part, but I found some issues.


Given that the attributes allow `0` as value we should allow the `VampirePlayer#getDbnoDuration` and the part in `Vampireplayer#getDbnoTimer` to allow working with 0 values instead of using `Math#max` to use a minimum of 1. The use case in `DBNOScreen#tick` should be changed accordingly. And we can change the min values of the config options `vpNeonatalDuration` and `vpDbnoDuration` to 0

src/main/java/de/teamlapen/vampirism/config/BalanceConfig.java
TheDrOfDoctoring: Oops, I was not thinking at all there :P I basically just thought that since it needs to be multiplied by a high positive number to increase the value it will have to be the opposite to get really low values, so I just put the negative of the max as the min, not thinking about how that would actually work.

Cheaterpaul:
I am not comfortable with adding an attribute that is not used by Vampirism itself.
However, I see your point that blood exhaustion can be low while blood levels are still full.

I recently increased the exhaustion for several vampire food items, which may change it a little bit.

Additionally I now added increased blood exhaustion for healing. This now uses the same parameters as the food stats healing. (No idea if vanilla changed or we changed these values).


After considering what the blood exhaustion actually does we can safely remove the new blood exhaustion modifier and replace all usages of `BloodStats#addExhaustion` that has the `ignoreModifier` parameter set to true with `false` or simply call the method without the parameter. That way it should do what you intended with the additional attribute.

Do you think that could work for you?

TheDrOfDoctoring:
Yeah, that would work for me, thanks!

Cheaterpaul (approved):

PR #1321, opened 2024-01-11
update to neoforge

Cheaterpaul:


maxanier:
Did you have to do all this manually, or is there some partial migration utility :O

Cheaterpaul:
There was an migration tool from 1.20.1 forge to 1.20.2 neoforge. But the Capability rework, network changes and registry changes as well as the "new" mnecraft changes too the most time.

PR #1406, opened 2024-08-16
Event for Vampire Fog Distance and Blood Food Mod Parity

TheGridExpert:
This adds a new event: `VampireFogEvent` which is fired after the `vampireBiomeFogDistanceMultiplier` is calculated in the RenderHandler. It allows add-on mods to change its value.

It also adds `getFoodProperties` to `VampirismItemBloodFoodItem` mostly for parity with other mods. If the player is a vampire it returns `vampireFood`, if not then the default food properties. AppleSkin mod uses getFoodProperties method to get the food value, so adding this will make it show the correct values for vampire players, like on this screenshot.
Screenshot 2024-08-16 at 19 10 46
Also, I added Properties parameter to `VampirismItemBloodFoodItem` constructor in case someone wants to extend this class.

(I also added in readme near my nickname that I also did structures in models, hope you don't mind)

### DIFF
--- a/src/api/java/de/teamlapen/vampirism/api/event/VampireFogEvent.java
+++ b/src/api/java/de/teamlapen/vampirism/api/event/VampireFogEvent.java
@@ -1,0 +1,37 @@
+package de.teamlapen.vampirism.api.event;
+
+import net.minecraft.client.player.LocalPlayer;
+import net.neoforged.bus.api.Event;
+import org.jetbrains.annotations.NotNull;
+
+@SuppressWarnings("unused")
+public class VampireFogEvent extends Event {
+    private final LocalPlayer player;
+    private float fogDistanceMultiplier;
+
+    public VampireFogEvent(@NotNull LocalPlayer player, float fogDistanceMultiplier) {
+        this.player = player;
+        this.fogDistanceMultiplier = fogDistanceMultiplier;
+    }
+
+    /**
+     * @return The Player that is seeing the fog.
+     */
+    @NotNull
+    public LocalPlayer getPlayer() {
+        return player;
+    }
+    /**
+     * @return The fog distance multiplier of vampire fog.
+     */
+    public float getFogDistanceMultiplier() {
+        return fogDistanceMultiplier;
+    }
+
+    /**
+     * @param fogDistanceMultiplier The fog distance multiplier of vampire fog
+     */
+    public void setFogDistanceMultiplier(float fogDistanceMultiplier) {
+        this.fogDistanceMultiplier = fogDistanceMultiplier;
+    }
+}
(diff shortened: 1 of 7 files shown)


src/main/java/de/teamlapen/vampirism/client/renderer/RenderHandler.java
```diff
@@ -160,6 +162,9 @@ public void onClientTick(ClientTickEvent.Pre event) {
                 vampireBiomeTicks--;
             }
         }
+
+        VampireFogEvent vampireFogEvent = VampirismEventFactory.fireVampireFogEvent(mc.player, vampireBiomeFogDistanceMultiplier);
+        vampireBiomeFogDistanceMultiplier = vampireFogEvent.getFogDistanceMultiplier();
```
Cheaterpaul: Is there a reason why this is at the end of the method and not at the same place as the multiplier is calculated at line 151?

src/api/java/de/teamlapen/vampirism/api/event/VampireFogEvent.java
```diff
@@ -0,0 +1,37 @@
...
+        this.fogDistanceMultiplier = fogDistanceMultiplier;
+    }
+
+    /**
+     * @return The Player that is seeing the fog.
+     */
+    @NotNull
+    public LocalPlayer getPlayer() {
+        return player;
+    }
```
Cheaterpaul: No need to add the player as parameter. We operate on the client here.

src/main/java/de/teamlapen/vampirism/util/VampirismEventFactory.java
```diff
@@ -97,4 +97,10 @@ public static ActionEvent.ActionUpdateEvent fireActionUpdateEvent(@NotNull IFact
         return event;
     }
 
+    public static VampireFogEvent fireVampireFogEvent(@NotNull LocalPlayer player, float fogDistanceMultiplier) {
+        VampireFogEvent event = new VampireFogEvent(player, fogDistanceMultiplier);
+        NeoForge.EVENT_BUS.post(event);
+        return event;
+    }
+
```
Cheaterpaul: We have a event factory for a reason, you can just return the modifier here, so that we have a cleaner code.

src/api/java/de/teamlapen/vampirism/api/event/VampireFogEvent.java
Cheaterpaul: If you want you can pass in the VampirePlayer

Cheaterpaul (changes requested):
Looks good, just a few minor changes

src/api/java/de/teamlapen/vampirism/api/event/VampireFogEvent.java
TheGridExpert: Makes sense, okay

Cheaterpaul:
:)

PR #1435, opened 2024-10-16
Update Repurposed Structures compatibility

TelepathicGrunt:
I did some refactoring in my template pool loading code and other file loading to be more like normal datapack loading without needing weird quirks like having the file be in same filepath as target pool  file.  Now the target pool/structure is specified within the json files here. Cleaner in long run.

There should be no more changes needed. Unless mojang changes template pools themselves but that seems super unlikely.

Also added house for cherry and bamboo villages.
image


Cheaterpaul:
Thanks for updating :rocket:

PR #1457, opened 2024-12-05
Blood vision: Indicate blood level

### LINKED ISSUE #1454: [Feature]: Blood vision upgrade that tells you how much blood the entity has
### What?

Well I would like a blood vision upgrade that tells you how much blood the entity has, a way I could see this implemented is that the red on a entity seen with blood vision will decrease the less blood it has. Preferably the top goes gray and that will just keep going down the less it has

### Mod experience

It will make it so that you dont have to check every entity individually after you come back to the same group

### Contributions

_No response_

maxanier:
Closes #1454
Scale blood vision entity color between red and gray depending on blood

The alpha modifier which we use has stopped working since 1.19. Still investigating whether this can be fixed


maxanier:
Probably not going to look further into the alpha modifier

PR #1481, opened 2025-03-07
Candelabras, chandeliers, villagers, maps

TheGridExpert:
### Crypt:
The building itself has been changed a bit, but now coffins and loot hanging from the floor are generated randomly. Some candles may be extinguished, brick cracks are now random and made using a processor list.

The underground part is also no longer flooded.

There's still a problem remaining though, because of terrain adaptation there are artefacts above the ground. It adapts the terrain as if the corridors were on top, signalling that there's something underground. Not sure how to fix that though, it seems to be a common issue for structures of such type.

### Village:
Hunter trainer houses have been updated.

Screenshot 2025-03-05 at 10 50 21
Screenshot 2025-03-05 at 11 02 28
Screenshot 2025-03-07 at 22 50 54
Screenshot 2025-03-05 at 10 53 55
Screenshot 2025-03-05 at 11 10 03

Representatives were given a new quest, Reopening Old Wounds, which gives a map that leads to ancient remains, so that players can just find them normally, rather than searching for ages only to find them buried underground (I've had this happen a couple of times where I managed to find the boss bar, but couldn't find the boss itself. It turned out to be about 7 blocks below the surface for some reason). WARNING: in this version, quests don't work in general, even without my changes, but it worked when I tested it on 1.21.1. Still will have to be tested again to make sure it works.

Villager progression is now as it should have always been. In vanilla, new villager trades always give more xp than the first level ones. In the mod, however, they all gave 2 xp, which made trading with the villagers a complete pain. Not only that, but they could only sell things, but not buy them, so you would need a lot of human hearts or vampire souls. Now they can buy more things from the player and give more xp.

The trades of the villagers have been changed in general with the main changes being:

- Both vampire and hunter experts now sell vampire forest map, previously only the vampire ones could, but the vampire ones can also sell crypt maps on level 4.
- Both can now sell ultimate weapons at level 5.
- Priests now also buy honeycombs and gold and sell crucifixes and candelabras.
- Non-vampire farmers can now buy garlic. Vampire butchers now buy hearts and masons buy and sell dark stone.
- Wandering traders were given new trades as well, they can now sell some overpriced stuff from vampire forest.

Screenshot 2025-02-28 at 10 16 05

### NPC AI
**[Requires discussion]** Hunter NPCs can now open gates. Also thought of improving their shooting AI since they currently just stand still and wait for a vampire to run to them.
They can now also walk on powder snow as they wear leather boots (feature from Vampire's Delight). Hunter players can also walk in case they have any hunter boots equipped.

**[Requires discussion]** Hunters, advanced hunters, vampires, and advanced vampires will now trigger when someone attacks their comrades. They'll also call for backup the same way if someone is injured by an enemy and the others haven't noticed the attacker yet. Vampires won't trigger if a vampire player attacks other vampires though.

### Other:
When a player tries to take blocks from a pillar, the block is now placed in the inventory instead of the hand.

Bonemealing cursed grass would not grow vampire orchids around anymore.

Added recipes for Bloody Dark Stone and Purple Stone Bricks

The ambient sound of Bloody Dark Stone Bricks is now split into single sounds of falling blood drops. Though there's a problem, sounds have echo which I didn't manage to get rid of.

`pack.mcmeta` file is now automatically datagenerated with the correct version number.

Tweaked the colours of spawn eggs and added new ones for hunter and vampire representatives as well as ghosts.

Blood bottles generated as dungeon loot are no longer empty which was caused by the remains of an old system with damage.

### Tags:
Added tag compat with Serene Seasons, garlic is now a summer and autumn crop. Also a few other tags, mostly for garlic.

Added the fireplace block to the `farmersdelight:tray_heat_sources' block tag for a little basic compatibility with Farmer's Delight (also a thing from Vampire's Delight, just thought it would be better to have it in the main mod in case the player doesn't install Vampire's Delight).

Vampire books and guide books can now be put into chiseled bookshelves (feature from Vampire's Delight)

### Remodels and Retextures:
Candelabras and chandeliers have been changed to match the candle stick.

Screenshot 2025-02-28 at 10 53 24
Screenshot 2025-02-28 at 10 56 55

New textures for cursed path has been added, it's now gray and goes better with other vampire biome blocks.

Screenshot 2025-02-28 at 10 33 12

Hunter intel textures have been updated to represent the book displayed on the hunter table.

Screenshot 2025-02-28 at 11 01 04

New Sanguinare effect texture has been added.

Screenshot 2025-03-07 at 10 06 37


src/main/java/de/teamlapen/vampirism/client/renderer/item/properties/HasName.java
```diff
@@ -0,0 +1,36 @@
...
+import net.minecraft.core.component.DataComponents;
+import net.minecraft.network.chat.Component;
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.item.ItemDisplayContext;
+import net.minecraft.world.item.ItemStack;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public record HasName(String name) implements ConditionalItemModelProperty {
```
Cheaterpaul, line 18: Is this class used somewhere?

src/main/java/de/teamlapen/vampirism/core/tags/ModEntityTags.java
```diff
@@ -13,6 +13,8 @@ public class ModEntityTags {
     public static final TagKey<EntityType<?>> ADVANCED_HUNTER = tag("advanced_hunter");
     public static final TagKey<EntityType<?>> ADVANCED_VAMPIRE = tag("advanced_vampire");
     public static final TagKey<EntityType<?>> CONVERTED_CREATURES = tag("converted_creatures");
+    public static final TagKey<EntityType<?>> COMRADE_DEFENDING_HUNTERS = tag("comrade_defending_hunters");
+    public static final TagKey<EntityType<?>> COMRADE_DEFENDING_VAMPIRES = tag("comrade_defending_vampires");
```
Cheaterpaul: I wouldn't name it comrade; I would name it something like alertable_hunters or something

src/main/java/de/teamlapen/vampirism/core/ModVillage.java
```diff
@@ -34,143 +40,264 @@ public class ModVillage {
...
+    public static final DeferredHolder<VillagerProfession, VillagerProfession> HUNTER_EXPERT = PROFESSIONS.register("hunter_expert", () -> new VillagerProfession(REFERENCE.MODID + ":hunter_expert", (holder) -> holder.is(ModPoiTypeTags.IS_HUNTER), (holder) -> holder.is(ModPoiTypeTags.IS_HUNTER), ImmutableSet.of(), ImmutableSet.of(ModBlocks.HUNTER_TABLE.get(), ModBlocks.WEAPON_TABLE.get(), ModBlocks.GARLIC.get()), SoundEvents.VILLAGER_WORK_ARMORER));
+    public static final DeferredHolder<VillagerProfession, VillagerProfession> PRIEST = PROFESSIONS.register("priest", () -> new VillagerProfession(REFERENCE.MODID + ":priest", holder -> ALTAR_CLEANSING.getKey() != null && holder.is(ALTAR_CLEANSING.getKey()), holder -> ALTAR_CLEANSING.getKey() != null && holder.is(ALTAR_CLEANSING.getKey()), ImmutableSet.of(), ImmutableSet.of(), ModSounds.BLESSING_MUSIC.get()));
```
Cheaterpaul: Is this null check necessary?

src/main/java/de/teamlapen/vampirism/data/provider/tags/ModBlockTagsProvider.java
```diff
@@ -166,8 +164,24 @@ protected void addTags(HolderLookup.Provider holderLookup) {
...
+            tag(ModBlockTags.STANDING_CANDELABRA).add(pair.getFirst());
+            tag(ModBlockTags.WALL_CANDELABRA).add(pair.getSecond());
+        });
+        tag(ModBlockTags.CANDELABRA).addTags(ModBlockTags.STANDING_CANDELABRA, ModBlockTags.WALL_CANDELABRA);
+        Helper.HANGING_CHANDELIERS.forEach(block -> tag(ModBlockTags.CHANDELIER).add(block));
+        tag(BlockTags.CANDLES).addTag(ModBlockTags.CANDLE_STICK).addTag(ModBlockTags.CANDELABRA).addTag(ModBlockTags.CHANDELIER);
+
+        tag(ModBlockTags.Compatibility.SERENE_SEASONS_AUTUMN_CROPS).add(ModBlocks.GARLIC.get());
+        tag(ModBlockTags.Compatibility.SERENE_SEASONS_SUMMER_CROPS).add(ModBlocks.GARLIC.get());
+        tag(ModBlockTags.Compatibility.TRAY_HEAT_SOURCES).add(ModBlocks.FIRE_PLACE.get());
```
Cheaterpaul, line 190: Can you put the compatibility tags in a separate `addCompatibilityTags` methods

src/main/java/de/teamlapen/vampirism/core/tags/ModBlockTags.java
```diff
@@ -39,4 +58,17 @@ public class ModBlockTags {
     private static @NotNull TagKey<Block> common(@NotNull String name) {
         return BlockTags.create(VResourceLocation.common(name));
     }
+
+    public static class Compatibility {
```
Cheaterpaul, line 62: good decision to make an extra class

src/main/java/de/teamlapen/vampirism/entity/minion/HunterMinionEntity.java
```diff
@@ -168,6 +169,7 @@ protected InteractionResult mobInteract(@NotNull Player player, @NotNull Interac
     @Override
     protected void registerGoals() {
         super.registerGoals();
+        this.goalSelector.addGoal(1, new OpenGateGoal(this, true));
```
Cheaterpaul: Not sure if I want minions to opens doors as well

src/main/java/de/teamlapen/vampirism/items/CandleHolderItem.java
```diff
@@ -0,0 +1,33 @@
...
+import net.minecraft.world.level.block.Block;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+public class CandleHolderItem extends BlockItem implements ModDisplayItemGenerator.CreativeTabItemProvider {
+    private final @Nullable List<CandleHolderBlock> allItems;
+
+    public CandleHolderItem(Block block, Properties properties, @Nullable List<CandleHolderBlock> allItems) {
```
Cheaterpaul: instead of providing all items as parameter or null; I would rather use the BlockItem and add the tab items in `VampirismCreativeTab`. Then you do not need to provide null as creative tab in the ModItem#register.

src/main/java/de/teamlapen/vampirism/blocks/DarkStoneBlock.java
```diff
@@ -1,41 +1,64 @@
...
+        if (state.is(ModBlocks.DARK_STONE_BRICKS) && stack.get(ModDataComponents.BOTTLE_BLOOD) != null) {
+            int blood = stack.getOrDefault(ModDataComponents.BOTTLE_BLOOD, new BottleBlood(0)).blood();
+
+            if (blood >= 3) {
+                level.setBlockAndUpdate(pos, ModBlocks.BLOODY_DARK_STONE_BRICKS.get().defaultBlockState());
+                level.playSound(null, pos, SoundEvents.SCULK_BLOCK_SPREAD, SoundSource.BLOCKS, 4.0F, 1.0F);
+
+                if (!player.hasInfiniteMaterials()) {
+                    stack.set(ModDataComponents.BOTTLE_BLOOD, new BottleBlood(blood - 3));
+                }
```
Cheaterpaul, line 53: return InteractionResult.SUCCESS_SERVER;

src/main/java/de/teamlapen/vampirism/items/CandleHolderItem.java
Cheaterpaul: The `CreativeTabProvider` is more used for providing additional sub items or items with different data components. But since we already have an item for every candle holder, like for most other items / blocks that just have a different variant, we should add the items to the create tab like any other one. But you can use the `Helper.HANGING_CHANDELIERS` etc there.

src/main/java/de/teamlapen/vampirism/items/PairCandleHolderItem.java
```diff
@@ -0,0 +1,35 @@
...
+import net.minecraft.world.level.block.Block;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+public class PairCandleHolderItem extends StandingAndWallBlockItem implements ModDisplayItemGenerator.CreativeTabItemProvider {
+    private final @Nullable List<Pair<CandleHolderBlock, CandleHolderBlock>> allItems;
+
+    public PairCandleHolderItem(Block block, Block wallBlock, Properties properties, @Nullable List<Pair<CandleHolderBlock, CandleHolderBlock>> allItems) {
```
Cheaterpaul: same here as in `CandleHolderItem`

src/main/java/de/teamlapen/vampirism/core/tags/ModDamageTypeTags.java
```diff
@@ -7,7 +7,6 @@
 import org.jetbrains.annotations.NotNull;
 
 public class ModDamageTypeTags {
-
```
Cheaterpaul, line 10: As general note for Pull Request, please refrain from doing cleanup on otherwise not referenced code. This just makes reviewing harder. Splitting this in a separate PR is totally fine.

src/main/java/de/teamlapen/vampirism/core/tags/ModItemTags.java
```diff
@@ -1,45 +1,56 @@
...
+
+    public static final TagKey<Item> HUNTER_INTEL = tag("hunter_intel");
     public static final TagKey<Item> VAMPIRE_BEACON_PAYMENT_ITEM = tag("vampire_beacon_payment_item");
-    public static final TagKey<Item> HEART_SEEKER = tag("swords/heart_seeker");
-    public static final TagKey<Item> HEART_STRIKER = tag("swords/heart_striker");
+
     public static final TagKey<Item> VAMPIRE_SLAYER_ITEMS = tag("enchantable/vampire_slayer");
     public static final TagKey<Item> CROSSBOW_ENCHANTABLE = tag("enchantable/crossbow");
 
+    public static final TagKey<Item> APPLICABLE_OIL_SWORD = tag("applicable_oil/sword");
```
Cheaterpaul, line 40: Same here for the whole file. Please refrain from doing such refactoring in the PR where you add or remove some fields. That makes it hard to find what you actually changed.

Cheaterpaul (changes requested):
Nice.

Looks mostly fine except for some suggestions and general notes. It looks like you already find your way through the vampirism code. 

Regarding the NPC AI @supermike1999 whats you opinion on these changes? For the gate change i would not allow hunter minion to open them.

Great idea to add the maps to the ancient remains and crypt.



I fix the conflicts later as promised.

Cheaterpaul:
And i think you forgot to add the new spawn eggs for the representatives and ghosts

TheGridExpert:
> And i think you forgot to add the new spawn eggs for the representatives and ghosts

Screenshot 2025-03-08 at 19 49 38

Weird, they are here.
Thank you. I will make the required changes (though there's one more thing I'm making right now, will commit them after it's done).

TheGridExpert:
Also, I forgot to mention this, but I've added new dripping ambient sounds for bloody dark bricks, thought they would sound better, but the echo present in them makes it sound weirder. I tried to get rid of it, but didn't manage to .

Cheaterpaul:
> Also, I forgot to mention this, but I've added new dripping ambient sounds for bloody dark bricks, thought they would sound better, but the echo present in them makes it sound weirder. I tried to get rid of it, but didn't manage to .

Yes the sound has some background noise there. Do you think [this](https://freesound.org/people/Millavsb/sounds/197900/) might fit better there?

Cheaterpaul:
> And i think you forgot to add the new spawn eggs for the representatives and ghosts

my bad already changed the branch when checking it

TheGridExpert:
> Yes the sound has some background noise there. Do you think [this](https://freesound.org/people/Millavsb/sounds/197900/) might fit better there?

Good, will put them there instead

TheGridExpert:
In [12c5a62](https://github.com/TeamLapen/Vampirism/pull/1481/commits/12c5a629b4630c06058172503f11deaa7b80953b) vampire books were basically rewritten, now they're stored in data, as the name suggests, so they can be added with data packs and other mods (Werewolves, for example) and translated. Also, books are now tagged and some structures have specific books. For example, hunter trainer houses now only have hunter books.
However, there's a potential thread to already-existing books. Their contents would just be deleted. I think it would make sense to add a method which is run when the game detects that the component is missing and finds a new component with the same id, although I am not sure how to implement that.

(That's what I meant by "though there's one more thing I'm making right now")

src/main/java/de/teamlapen/vampirism/world/loot/functions/SetVampireBookFunction.java
```diff
@@ -0,0 +1,87 @@
...
+                ResourceLocation id = ResourceLocation.parse(provider.getBookLootId().get());
+                Optional<IVampireBook> possibleBook = registryAccess.lookupOrThrow(VampirismRegistries.Keys.VAMPIRE_BOOK).getOptional(id);
+
+                if (possibleBook.isPresent()) {
+                    vampireBook = possibleBook.get();
+                } else {
+                    LOGGER.warn("Vampire Book \"{}\" does not exist, cannot add it to a loot table", id.getPath());
+                }
+            } else {
+                TagKey<IVampireBook> factionTag = Helper.isVampire(entity) ? ModVampireBookTags.IS_VAMPIRE : Helper.isHunter(entity) ? ModVampireBookTags.IS_HUNTER : ModVampireBookTags.IS_GENERAL;
```
Cheaterpaul: When we want different faction to use book then register the faction tags to the faction in the `FactionBuilder` in `ModFactions` and retrieve the tag using `IFaction#getTag` with the general tag as fallback.

src/main/resources/supporters.json
```diff
@@ -215,7 +215,7 @@
     "texture": "Volqonah",
     "type": 15,
     "status": "Temporary",
-    "bookId": "volqonah",
+    "bookId": "vampirism:valorous_tale",
```
Cheaterpaul: this is tricky, because the supporter file is uploaded and downloaded for older versions as well. This means that it must be backwards compatible. The mod gets the current supporter from [the api](https://api.vampirism.dev/api/v1/supporter/list). Therefor we can not use the same field for different ids.

We can name the new field just "book". In the code we then can ignore the "bookId" but in the supporter.json the "bookId" must be kept with the old values while adding the new key pairs as well.

src/main/resources/assets/vampirism/lang/en_us.json
```diff
@@ -1608,6 +1609,71 @@
   "stat_type.vampirism.action_cooldown_time": "Action On Cooldown",
   "stat_type.vampirism.skill_unlocked": "Skill Unlocked",
   "stat_type.vampirism.skill_forgotten": "Skill Forgotten",
+  "__comment": "vampire books",
+  "vampire_book.vampirism.unknown_id": "Unknown Book",
```
Cheaterpaul: I do not feel good about adding the book content in the lang files as discussed previously. Lang files are supposed to contain a simple list of key value pairs to describe things.


They are not a place to store load amount of book content. [Here](https://github.com/neoforged/NeoForge/issues/1519#issuecomment-2337833818) was a little bit a discussion about a similar thing.


Regarding the translation, I doubt that there are a lot of people who want to translate this correctly into the languages and not just use an AI to do this. Then we do not need to put it in a languages file and publish it to crowdin which has word limits.

Cheaterpaul (changes requested):
I agree we can do the data gen approach to make it easier to categorize and add books, but the whole "put the book content in the lang files" is bothering me. 

We can not put loads of text in the lang file. That is simply not what they should be used for. (yeah i know the guide book does it, but that is a whole other topic)

I agree there is some need to have them translated, but i don't know to what extend. But my thought for a solution to be discussed:
- keep the title translation in the lang file
- add a author translation that uses formatting: "by %s" for english and fill in the author name from the datapack entry
- put the text into separate files (not the lang files) that we load by our self and where we manage the translations in git only
  - we could have some `assets/vampirism/vampire_book` folder to store the different books in different subfolders for each language to manage translations
  - translations might be added to the mc translations map to still use `Component`

TheGridExpert:
Okay. It's going to be tricky, but I will try to do something with all of that

TheGridExpert:
> put the text into separate files (not the lang files) that we load by our self and where we manage the translations in git only
we could have some assets/vampirism/vampire_book folder to store the different books in different subfolders for each language to manage translations
translations might be added to the mc translations map to still use Component

Did I understand correctly that there is a folder vampire_book and there are folders with book names inside, and inside of them there are translation files for each language?

Cheaterpaul:
i would suggest something like:
- assets
  - vampirism
    - vampire_book
      - my_prince
        - en_us.json
        - de_de.json
      - dear_martha
        - en_us.json
        - de_de.json

TheGridExpert:
I've made the most of the changes regarding vampire books. Currently, the book text is stored in json and loaded when the player opens a book. Though I'm not sure if you meant the same to be done with author names or not. 

>add a author translation that uses formatting: "by %s" for english and fill in the author name **from the datapack entry**

Should the names be untranslatable? There could be a potential problem with that, I think. The initial reason why I put "by" into all the translation lines is that it may result in some languages like Ukrainian or Russian looking broken. The "by" part is translated separately, meaning that it would look weird with one part being in Latin and another one in Cyrillic. It would be even worse with the "Unknown" author because it would look like "автор: Unknown" which works in English, but here it's put as if it was the name. If it is translated though, then the "by" part shouldn't be separated or just have a separate line for the "Unknown" one.

src/main/java/de/teamlapen/vampirism/util/VampireBookLoader.java
```diff
@@ -0,0 +1,63 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+
+public class VampireBookLoader {
+
+    private static final Logger LOGGER = LogManager.getLogger();
+
+    public static List<String> loadBookContents(IVampireBook vampireBook) {
+        String languageCode = Minecraft.getInstance().getLanguageManager().getSelected();
```
Cheaterpaul: Can we put this in a method in the proxy to remove the dependency on the Minecraft class here. That way there is no problem if a vampire book is loaded on the server.

Just create a getLanguage method in the IProxy and return en_us in the default implementation, while returning the client language in the ClientProxy

src/main/java/de/teamlapen/vampirism/util/VampireBookLoader.java
```diff
@@ -0,0 +1,63 @@
...
+        List<String> contents = loadFile(path);
+        return contents.isEmpty() ? List.of("§4Failed to load the contents file: " + path + "§r") : contents;
+    }
+
+    private static String getFilePath(IVampireBook vampireBook, String languageCode) {
+        return "assets/" + vampireBook.id().getNamespace() + "/vampire_books/" + vampireBook.id().getPath() + "/" + languageCode +".json";
+    }
+
+    private static boolean fileExists(String path) {
+        return Thread.currentThread().getContextClassLoader().getResource(path) != null;
```
Cheaterpaul: Just use `VampirismMod.class.getResource(path)` that should do the same and skips the thread looking for the class loader that the class already known about.

src/main/java/de/teamlapen/vampirism/util/VampireBookLoader.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    private static List<String> loadFile(String path) {
+        final ClassLoader classLoader = Thread.currentThread().getContextClassLoader();
+        List<String> lines = new ArrayList<>();
+
+        try (InputStream inputStream = classLoader.getResourceAsStream(path)) {
```
Cheaterpaul: same here

src/main/java/de/teamlapen/vampirism/items/component/VampireBook.java
```diff
@@ -0,0 +1,84 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.tags.TagKey;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.storage.loot.LootContext;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.stream.Stream;
+
+public record VampireBook(ResourceLocation id, ResourceLocation itemModel, ResourceLocation backgroundTexture) implements IVampireBook {
```
Cheaterpaul: Is the item model used somewhere?

Cheaterpaul:
```suggestion
public record VampireBook(ResourceLocation id, ResourceLocation itemModel, ResourceLocation backgroundTexture, Component author) implements IVampireBook {
```

src/main/java/de/teamlapen/vampirism/items/component/VampireBook.java
```diff
@@ -0,0 +1,84 @@
...
+
+    private static final VampireBook EMPTY = new VampireBook(VResourceLocation.mod("unknown_id"), ModVampireBooks.DEFAULT_ITEM_MODEL, ModVampireBooks.DEFAULT_BACKGROUND_TEXTURE);
```
Cheaterpaul:
```suggestion
    private static final VampireBook EMPTY = new VampireBook(VResourceLocation.mod("unknown_id"), ModVampireBooks.DEFAULT_ITEM_MODEL, ModVampireBooks.DEFAULT_BACKGROUND_TEXTURE, Component.translatable("text.vampirism.unknown"));
```

src/main/java/de/teamlapen/vampirism/items/component/VampireBook.java
```diff
@@ -0,0 +1,84 @@
...
+
+    public static final Codec<IVampireBook> CODEC = RecordCodecBuilder.create(instance -> instance.group(
+            ResourceLocation.CODEC.fieldOf("id").forGetter(IVampireBook::id),
+            ResourceLocation.CODEC.optionalFieldOf("item_model", ModVampireBooks.DEFAULT_ITEM_MODEL).forGetter(IVampireBook::itemModel),
+            ResourceLocation.CODEC.optionalFieldOf("background_texture", ModVampireBooks.DEFAULT_BACKGROUND_TEXTURE).forGetter(IVampireBook::backgroundTexture)
+            ).apply(instance, VampireBook::new)
```
Cheaterpaul:
```suggestion
    public static final Codec<IVampireBook> CODEC = RecordCodecBuilder.create(instance -> instance.group(
            ResourceLocation.CODEC.fieldOf("id").forGetter(IVampireBook::id),
            ResourceLocation.CODEC.optionalFieldOf("item_model", ModVampireBooks.DEFAULT_ITEM_MODEL).forGetter(IVampireBook::itemModel),
            ResourceLocation.CODEC.optionalFieldOf("background_texture", ModVampireBooks.DEFAULT_BACKGROUND_TEXTURE).forGetter(IVampireBook::backgroundTexture),
            ComponentSerialization.CODEC.optionalFieldOf("author", EMPTY.author()).forGetter(IVampireBook::author)
            ).apply(instance, VampireBook::new)
```

Cheaterpaul (changes requested):
Very nice solution for the vampire books. :+1: Look all fine to me.

For the author names i thought we just have the names, but as you said we have some "Unknown" and "Unknown Priest" authors there which require translations. But for the actual names we do not want to translate the actual names of the authors.

I a solution i would this we could to the following:
- Add the authors to the data pack object so that is is loaded in the vampire book object if present.
- The author field is loaded as optional `Component` (see `ComponentSerialization#CODEC`)
- When we have an author store it as `Component#literal`, if we have no author ("Unknown") keep it empty and add a "Unknown" translatable component to the EMPTY book (the optional codec should keep the unknown author out of the datapack)
- Add a translatable component to the book that has "Unknown Scribe" as author

(I added some diffs to explain a little bit better what i meant, they don't compile and work)

What do you think. Does that make sense?

Cheaterpaul:
I talked with mike and he had 3 points:
1. the alertable thing is good, but for lore reasons it would make sense for all hunters alert each other, but for vampires it does not make sense for the basic vampires to alert others. Vampires are generally not working together unless "forced". SO it would make sense for advanced vampires to alert nearby vampires, but not for the basic ones to alert other.

I agree with him there.

2. Why should the vampire orchids no longer spawn when using bone meal?

3. The Ancient Remains task reward must be expensive, very rare and limited. Yes it is annoying to search for it, but that is somewhat the point. You should not simply try to cheese for the task because it is so much easier than searching in the world for it.

I agree as well. If it is to easy to get and complete the task, then there is no reason to try and find it on another way. So either we need to make it expensive enough to make people think: "Damn that is to complicated/expensive, i try my luck and search for it in the world" or, my preferred way, we remove the task and make it a loot item that can be found "vary rarely" in some structure (for example the crypt). That would bring people more reasons to explore and reword the exploring.

What do you thing?

Otherwise he liked the changes.

src/main/java/de/teamlapen/vampirism/items/component/VampireBook.java
TheGridExpert: Both itemModel and backgroundTexture were supposed to be used, but for now I've kept it like this as I wasn't sure whether implementing custom ones wouldn't make it even more difficult for players to understand that it's a vampire book. Well, the main reason though is that I simply didn't manage to get those custom models/textures to bake. Those just become invisible and nothing helps.

TheGridExpert:
> 1. the alertable thing is good, but for lore reasons it would make sense for all hunters alert each other, but for vampires it does not make sense for the basic vampires to alert others. Vampires are generally not working together unless "forced". SO it would make sense for advanced vampires to alert nearby vampires, but not for the basic ones to alert other.

Yep, makes sense. I also see hunters as more cooperative and human (that's also why they can open gates and vampires can't) Will add.

> 2. Why should the vampire orchids no longer spawn when using bone meal?

Now I'm not sure about that decision, but there were two reasons:
1. There's a villager trade for vampire orchids which used to be 4 flowers for an emerald. If someone made a flower farm on bonemeal it would be cheesy. It was changed to 8 pr emerald later when I changed trades, but still unsure.
2. The same thing about the bone meal farms, but more personal, I would say. When I added vampire orchids as crops to Vampire's Delight I intended those to be farmed like that and not through bone meal farms which are insane in terms of flower amount.

> 3. The Ancient Remains task reward must be expensive, very rare and limited. Yes it is annoying to search for it, but that is somewhat the point. You should not simply try to cheese for the task because it is so much easier than searching in the world for it.

> I agree as well. If it is to easy to get and complete the task, then there is no reason to try and find it on another way. So either we need to make it expensive enough to make people think: "Damn that is to complicated/expensive, i try my luck and search for it in the world" or, my preferred way, we remove the task and make it a loot item that can be found "vary rarely" in some structure (for example the crypt). That would bring people more reasons to explore and reword the exploring.

To be honest, the idea of making people unable to purchase the map because of the cost and forcing them to find those in the wild doesn't seem too good. Those are really rare, just making people "explore the world" would be sadistic while an alternative exists (and it just comes to endless walking from one side of the universe to another as they are simply rare as hell, sorry for my English. And eventually it just comes to you trying to find a chopped down tree for a month only to find a boss bar and no boss because it spawned underground, it often occurs, which is something the map also helps with because you would know where to dig). (Also some players have argued that the reward from boss itself doesn't match the efforts, though can't say for sure, I have managed to find ancient remains only once by myself).

I agree that it could spawn in a structure, though the initial idea about making it a task was that this "mission" would be given to you by your higher-ups through a task master. Would make sense especially for hunters ("Hey, we've noticed you're doing well, go fight a mother, comrade, we need your help with that"). Also new players may discover the existence of a boss through the tasks, not everyone looks into the wiki to find information about the mod. Many just jump in and understand as it goes or just look for a guide on YouTube which are all outdated. Maybe making those spawn in crypts, why not (though it would also mean GRIND, GRIND, GRIND but now for villagers and stuff to level-up them to level 4 so that they sell crypt maps, those things are also rare. Maybe making a separate structure for those maps or just make them a little more common than "vary rarely?).

TheGridExpert:
I have made the changes. Now the author names are stored as components in data. Some are untranslatable, some are not.



(That's the last new thing I add in this pull request, I swear)
Also changed the cloaks. They now overlay the player capes, though now it only works with players and not with mobs. Initially I thought of adding a separate slot in Vampirism menu, below the refinements, so that you could equip them together with armour, but was neither sure how to do that nor that it's a good idea to make it this way. Added more color variants as well.

PS: Some of the cloak textures are temporary, especially the orange and the yellow ones. I will change them when I get how to make orange look like orange.

src/generated/resources/data/vampirism/vampirism/vampire_book/royal_rivalry.json
```diff
@@ -1,3 +1,7 @@
 {
+  "author": {
+    "translate": "vampire_book.vampirism.royal_rivalry.author"
+  },
+  "hasByPreposition": false,
```
Cheaterpaul: run the data client gen once more

src/main/resources/assets/vampirism/lang/en_us.json
```diff
@@ -76,7 +76,7 @@
   "text.vampirism.lord.portal_disabled": "This portal does not lead anywhere",
```
Cheaterpaul: Did you maybe replaced to many `by`s ? I see that every `by` is removed

src/main/java/de/teamlapen/vampirism/mixin/client/AbstractClientPlayerMixin.java
```diff
@@ -0,0 +1,26 @@
...
+        AbstractClientPlayer player = (AbstractClientPlayer) (Object) this;
+        Item item = player.getItemBySlot(EquipmentSlot.CHEST).getItem();
+
+        if (item instanceof ICapeItem capeItem) {
+            return new PlayerSkin(original.texture(), original.textureUrl(), capeItem.getCapeTexture(), original.elytraTexture(), original.model(), original.secure());
+        }
+        
+        return original;
+    }
+}
```
Cheaterpaul: That part it the one that makes problems.

Mojang explicitly forbids to add custom capes and the use if you do not own a cape.
The Neoforge developers discourage to go near that and do something with that, so I want to stay away from that as well.
That means that we can not use the cape for the cloaks and need to use an own model for it. The model may look similar like the current models, and behave similar, but cannot be the cap model.

Cheaterpaul (changes requested):

src/generated/resources/data/vampirism/vampirism/vampire_book/royal_rivalry.json
TheGridExpert: Yes, I forgot for some reason

src/main/java/de/teamlapen/vampirism/mixin/client/AbstractClientPlayerMixin.java
TheGridExpert: That's bad... But how do mods like Capes, ETF or Essential work though? They do overlay your cape, you can even choose your own one. I think they do prohibit selling those or something, but not putting your own ones for free

Cheaterpaul:
> > 2. Why should the vampire orchids no longer spawn when using bone meal?
> 
> Now I'm not sure about that decision, but there were two reasons:
> 
>     1. There's a villager trade for vampire orchids which used to be 4 flowers for an emerald. If someone made a flower farm on bonemeal it would be cheesy. It was changed to 8 pr emerald later when I changed trades, but still unsure.
> 
>     2. The same thing about the bone meal farms, but more personal, I would say. When I added vampire orchids as crops to Vampire's Delight I intended those to be farmed like that and not through bone meal farms which are insane in terms of flower amount.
> 

That shouldn't be a reason to remove them from the bone meal effect.
Sugar cane trades for example are very cheese as well and i don't doubt that there are others as well. Then i would just make the trade even more expensive 16 orchids for 1 emerald. Then you still need to spend bone meal. The sugar cane just grows.

The spawn chance for orchids seems to be somewhat 2 per bone meal, which we can reduce as well (if that is possible).





> > 3. The Ancient Remains task reward must be expensive, very rare and limited. Yes it is annoying to search for it, but that is somewhat the point. You should not simply try to cheese for the task because it is so much easier than searching in the world for it.
> 
> > I agree as well. If it is to easy to get and complete the task, then there is no reason to try and find it on another way. So either we need to make it expensive enough to make people think: "Damn that is to complicated/expensive, i try my luck and search for it in the world" or, my preferred way, we remove the task and make it a loot item that can be found "vary rarely" in some structure (for example the crypt). That would bring people more reasons to explore and reword the exploring.
> 
> To be honest, the idea of making people unable to purchase the map because of the cost and forcing them to find those in the wild doesn't seem too good. Those are really rare, just making people "explore the world" would be sadistic while an alternative exists (and it just comes to endless walking from one side of the universe to another as they are simply rare as hell, sorry for my English. And eventually it just comes to you trying to find a chopped down tree for a month only to find a boss bar and no boss because it spawned underground, it often occurs, which is something the map also helps with because you would know where to dig). (Also some players have argued that the reward from boss itself doesn't match the efforts, though can't say for sure, I have managed to find ancient remains only once by myself).
> 
> I agree that it could spawn in a structure, though the initial idea about making it a task was that this "mission" would be given to you by your higher-ups through a task master. Would make sense especially for hunters ("Hey, we've noticed you're doing well, go fight a mother, comrade, we need your help with that"). Also new players may discover the existence of a boss through the tasks, not everyone looks into the wiki to find information about the mod. Many just jump in and understand as it goes or just look for a guide on YouTube which are all outdated. Maybe making those spawn in crypts, why not (though it would also mean GRIND, GRIND, GRIND but now for villagers and stuff to level-up them to level 4 so that they sell crypt maps, those things are also rare. Maybe making a separate structure for those maps or just make them a little more common than "vary rarely?).

The problem with the tasks is that you can refresh your task list by killing the representative and wait for a new one or just wait a day. That way it is relatively easy to get the task and a guarantee at finding it. You can do that at home and do not need to travel, besides farming the resources or completing the quest.

The ancient remains should be rare, maybe not as rare as it is now, but rare. I see no problem in people leveling up the villager to get the crypt map, just like any other villager.

We also need to keep in mind that we can not create many maps for the same ancient remains, because every map is useless once the ancient remains is destroyed.
Yes, the crypt is "rare" as well, bus still common in comparison, so i see no problem in letting the map only spawn there. If you want even in the altar structure or a new fitting structure. But they should not be much more common than the current structures.

But we can increase the spawn chance of the ancient remains by ~100% if that is doable with the settings.

TheGridExpert:
Reasonable, I think. I think it could be done this way. Though 100% would be too much, probably a bit less. (Also yes, crypt isn't too rare compared to others). From all the solutions, the crypt is the best-fitting one for now.

> We also need to keep in mind that we can not create many maps for the same ancient remains, because every map is useless once the ancient remains is destroyed.

Wait, so when the remains are destroyed, the structure technically disappears from the world? So, technically, if you kill the vampire representative and get a new one in the same village, he would sell you a different map? If yes, it would be more logical to put the map into the crypt, yes. Will do it in the near time.

> That shouldn't be a reason to remove them from the bone meal effect.
Sugar cane trades for example are very cheese as well and i don't doubt that there are others as well. Then i would just make the trade even more expensive 16 orchids for 1 emerald. Then you still need to spend bone meal. The sugar cane just grows.

> The spawn chance for orchids seems to be somewhat 2 per bone meal, which we can reduce as well (if that is possible).

Makes sense, I will revert it.

Cheaterpaul:
Although the blocks are destroyed, the "structure" which is registered and placed though the world's seed stays, because the structure is not an object in the world (Although the chunks have a stored reference to the structure). But generally speaking a structure is where i can calculate that a structure is (with seed and spawn settings) no matter what happened to the world there. So even a trader can sell me a map to an already raided structure.

TheGridExpert:
Screenshot 2025-03-15 at 23 52 33
I've researched about the capes. It seems that the only place where EULA, usage guidelines, and other policies mention specifically capes is when it comes to selling them. They outline that you can't charge money for them. Pretty much the reason why Essential has a whole in-game cosmetic shop, but all capes are given for free and they're usable. As I've mentioned before, there are plenty of mods which do change them for free and they're popular enough for it to get noticed. Quark, for example, has custom capes for developers. Even if this did apply on any cape code modification, the mod would have already been in trouble because of the part "anything that attempts to visually act like the feature" (meaning cape), so there's not a big difference between making the cape as a separate model or overlaying the existing one.

Cheaterpaul:
Well we can the following
- What i get from the communication of the neoforge developer and contributes is that mojang does not want mods to modify the capes. Is is not clear what they actually do not want. Thus I do not necessarily want to go into that direction.
- Just because other mods do this does not mean we should.
- But the more important thing is that we have a cloak item that should be noticeable as cloak in game. Vampires wear cloaks and as such the item should reassemble a cloak. The baron cloak nicely shows how it should look like. The player cloak item more or less looks like an easy solution to the problem to design and implement a actual cloak. Switching to a cape as model, would change the meaning of the whole item, when it is supposed to be a cloak.

Yes capes are fancy, but do we need a thing, just because it is fancy. The Vampire wings are requested by player since the beginning but are yet to be added.

TheGridExpert:
Makes sense, okay. I will bring back the old model. Though would it be okay if I made it so it disables the cape from showing while wearing the cloak since it looks quite odd when those two overlap? (Not sure, but maybe it would also be considered as modifying)

Cheaterpaul:
thats fine with me

TheGridExpert:
Sorry for the delay. I've changed the cloaks. Now they work as a model as they did before. Their animation is different from the cape though, it's less moving, as you can see on the video. There's also a cloak holder around the player's neck, took the idea from the original cloaks. I hope they're okay now.

https://github.com/user-attachments/assets/78ef3cff-408f-40f4-8104-6acff063dfe8

About the ancient remains. I moved the map from the task to the crypt, as discussed. The chance of finding it is a bit more than 5%. It would require to loot from two to three crypts to find one on average. I did not change the ancient remains spawn chance as I wasn't sure about the exact number, so left it to you /:

Changed advancements as well. There are now three new advancements. One is for finding a crypt, it's called "Domain of the Dead" (thanks to Random for the name idea). Another one is for finding a map in the crypt which is "Reopening Old Wounds" (so original, surely didn't use it earlier) and "Plague Inc." for infecting a mob as a vampire. Besides that, there have also been a few changes the icons and the description of some of the advancements. Changed the icon of "Blood Feeder's Home" to a dark spruce sapling (it used to be an oak log), an advancement for becoming a lord now has a crown and not a paper, "Resource Collector" now has a bundle, and "The Guardian" was given a red banner because there were two advancements with a shield in the same branch. As well as that, blood bottles and containers are now filled with blood on the icons. As for the descriptions, just made some of them more clear and removed the duplicate of text which was in one of those.

Screenshot 2025-03-27 at 21 51 44

Screenshot 2025-03-27 at 21 51 23

Screenshot 2025-03-27 at 21 51 31

Not only that, but also changed the vampire soul lantern model. Took it from Vampire's Delight since there was already such a block there, why not. Though faced one issue. As you can see, I've added creeper_repellent tag for items as well. I initially intended to make it so creepers avoid you while holding the lantern, not only when it's placed, but then realised that they don't avoid lanterns in general for some reason. I wasn't sure why, but it looks like it a custom goal just doesn't get applied (maybe the same reason why golems attack villagers). So I just decided not to touch it. Also changed its recipe to use golden ingots and a vampire soul as it's made of gold.



As for the technical, I've renamed BushBlock to CursedRootsBlock to avoid confusion with the vanilla block class. (It's an often case in general for many other classes, especially data gen. I've sometimes used the wrong one while developing Vampire's Delight (especially the BushBlock class), it may be confusing even inside the Vampirism itself).

A small change, but vampire orchids now poison bees. It's a new function added in Minecraft recently for toxic flowers such as wither rose and the one from the pale garden. Also removed the type enum from the class as it can be done more easily, just with an effect in the constructor itself.

Added the sounds you offered earlier to be used as the blood drip ones.

Bonemealing cursed grass now both generates vampire orchids and cursed roots.

Added a few more vampier book translations.

Advanced vampires no longer help their lower-tier brothers.


I hope I haven't added even more work to remake for myself.

src/main/resources/assets/vampirism/lang/en_us.json
```diff
@@ -225,57 +227,63 @@
...
+  "advancement.vampirism.max_level_hunter.desc": "Reach the highest hunter level",
   "advancement.vampirism.technology": "Modern Technology",
   "advancement.vampirism.technology.desc": "Get yourself an Advanced Crossbow",
   "advancement.vampirism.extra_storage": "Blood Bank",
-  "advancement.vampirism.extra_storage.desc": "Get yourself a Blood Container for all that Blood",
+  "advancement.vampirism.extra_storage.desc": "Get yourself a Blood Container for all that blood",
   "advancement.vampirism.ancient_knowledge": "Vampire Diaries",
   "advancement.vampirism.ancient_knowledge.desc": "Find a Vampire Book",
+  "advancement.vampirism.domain_of_the_dead": "Domain of the Dead",
+  "advancement.vampirism.domain_of_the_dead.desc": "Follow the map from the Vampire Expert and find the resting place of the dead",
```
Cheaterpaul: I would not write the vampire expert into the advancement description, unless it request the map from a vampire expert to trigger it. But you can find the crypt without one, so i would leave it out of here.

src/main/resources/assets/vampirism/lang/en_us.json
```diff
@@ -74,7 +76,7 @@
   "text.vampirism.lord.portal_disabled": "This portal does not lead anywhere",
   "text.vampirism.lord.level_to_low": "You vampire level is too low to be a Vampire Lord",
   "text.vampirism.coffin.wrong_dimension": "You can't sleep in this dimension.",
-  "text.vampirism.deactivated_by_serveradmin": "This is deactivated by your server admin",
+  "text.vampirism.deactivated_by_serveradmin": "This is deactivated your server admin",
```
Cheaterpaul: Still the issue about the `by`s that were removed in 7199a5d4f1151b5f1c992902dca0c0945e80dad7

Cheaterpaul (changes requested):
Looks good :+1:

Just these small things

TheGridExpert:
Fixed the "by"s (forgot about it previously) and changed the description of the advancement to "Stumble upon a crypt, the place where the undead rest", I suppose it would fit better.

Also, as you can see, renamed hunter hats from "0" to "1" to "tall" and "broad" to make it easier to distinguish them.

src/main/java/de/teamlapen/vampirism/entity/hunter/HunterBaseEntity.java
```diff
@@ -120,8 +120,8 @@ public ItemStack getOffHand() {
 
     public enum HatType {
         NONE(null),
-        HAT_0(ModItems.HUNTER_HAT_HEAD_0),
-        HAT_1(ModItems.HUNTER_HAT_HEAD_1);
+        TALL(ModItems.HUNTER_HAT_TALL),
```
Cheaterpaul, line 123: renaming the enum values causes the HatType#get method to no longer finds the correct hat when loading the advanced hunter hat type from the Supporter `AdvancedHunterEntity#applyCustomisationItems`, because the supporter have a `hat_0` or `hat_1` value set.

Please map the old values to the new values in the `HatType#get` method

Cheaterpaul: Alternatively we add a new "headwear" values to the appearance map, that will use the new hat strings and allows more customization. We could even use the registry id of the item in the appearance map instead of relying on the hat type.

TheGridExpert:
Something like this?

src/main/java/de/teamlapen/vampirism/entity/hunter/HunterBaseEntity.java
TheGridExpert: Yes, that would be better. I will change it to this

TheGridExpert: (I will do it, but tomorrow. It's quite late here already)

TheGridExpert:
Now it parses the headwear line to get the hat. If it's air, then the hat is missing. If it cannot be parsed or the line does not exist, it uses a random hat.



(Set Rand's hat to an iron helmet just to test. It's the same as it used to be)

Screenshot 2025-03-29 at 11 48 31

By the way, while I was changing the hunter hats, I noticed that there's an easter egg left. It was used when Vampirism reached 10 000 000 downloads, if I'm not mistaken, but now it ended and it's left unused. Is there any point in keeping it in the class? (That's basically the only non-constructor method in the class).

```java
public class HunterHatItem extends HunterArmorItem {

    public HunterHatItem(ArmorMaterial armorMaterial, Item.Properties properties) {
        super(armorMaterial, ArmorType.HELMET, properties);
    }

    @Override
    public void inventoryTick(ItemStack pStack, Level pLevel, Entity pEntity, int pSlotId, boolean pIsSelected) {
        super.inventoryTick(pStack, pLevel, pEntity, pSlotId, pIsSelected);
        if (pSlotId >= 36 && pSlotId <= 39 && pEntity instanceof Player living) {
            if (pStack.has(DataComponents.CUSTOM_NAME) && "10000000".equals(pStack.getHoverName().getString()) && VampirismAPI.settings().isSettingTrue("vampirism:10000000d")) {
                UtilLib.spawnParticlesAroundEntity(living, ParticleTypes.ELECTRIC_SPARK, 0.5, 4);
                if (living.tickCount % 16 == 4) {
                    living.addEffect(new MobEffectInstance(MobEffects.LEVITATION, 30, 0));
                    living.addEffect(new MobEffectInstance(MobEffects.SLOW_FALLING, 100, 2));
                }
            }
        }
    }
}

File: src/main/java/de/teamlapen/vampirism/entity/hunter/AdvancedHunterEntity.java [hunk 27]
```diff
@@ -334,11 +334,13 @@ private void applyCustomisationItems(Supporter supporter) {
         this.setItemSlot(EquipmentSlot.MAINHAND, equipment.getMainHand());
         this.setItemSlot(EquipmentSlot.OFFHAND, equipment.getOffHand());
 
-        HatType hat = Optional.ofNullable(appearance.get("hat")).map(HatType::get).orElseGet(() -> {
+        ResourceLocation headwearId = ResourceLocation.parse(appearance.get("headwear"));
```
Cheaterpaul: the headwear might not be set or might be invalid so we need to first check for null and catch the parse exception that might be thrown.

If the parsing fails or the item is not present, we should not assign an item and log a warning.
If none is set we should assign a random one.

TheGridExpert:
The initial reason why I made it this way was because DefaultedRegistry#getValue(ResourceLocation resourceLocation) (which BuiltInRegistries.ITEM is) returned Items.AIR in any way, whether it was actually air or it failed to parse. Found out that the same method but with a ResourceKey actually returns null if the parse failed. Changed it to use it, now it should work properly even if the field is missing.

src/main/java/de/teamlapen/vampirism/entity/hunter/AdvancedHunterEntity.java
```diff
@@ -334,13 +335,19 @@ private void applyCustomisationItems(Supporter supporter) {
...
+
+        Item headwearItem = headwear == null ? randomHat : BuiltInRegistries.ITEM.getValue(ResourceKey.create(BuiltInRegistries.ITEM.key(), ResourceLocation.parse(headwear)));
+        if (headwearItem == null) {
+            headwearItem = randomHat;
+            LogUtils.getLogger().warn("Failed to parse the id \"{}\" of advanced hunter {}'s headwear", headwear, supporter.name());
         }
-        this.setItemSlot(EquipmentSlot.HEAD, headwear.getDefaultInstance());
+
+        this.setItemSlot(EquipmentSlot.HEAD, headwearItem.getDefaultInstance());
+
```
Cheaterpaul: I was thinking more like this:

```java
        String headwear = appearance.get("headwear");
        Item headWearItem = null;

        if (headwear != null) {
            ResourceLocation headWearId = ResourceLocation.tryParse(headwear);
            if (headWearId == null) {
                LogUtils.getLogger().warn("Failed to parse the id \"{}\" of advanced hunter {}'s headwear", headwear, supporter.name());
            } else {
                headWearItem = BuiltInRegistries.ITEM.getValue(ResourceKey.create(Registries.ITEM, headWearId));
            }
        }

        if (headWearItem == null) {
            HatType[] types = HatType.values();
            headWearItem = types[random.nextInt(types.length)].getHeadItem().getItem();
        }

        this.setItemSlot(EquipmentSlot.HEAD, headWearItem.getDefaultInstance());
```

The important point i wanted to make is that `ResourcesLocation#parse` can throw an `ResourceLocationException` when the string has an invalid format. `ResourceLocation#tryParse` checks the format. Because we get the string from an external source we need to check this. (Even when we ourselves are the external source)

TheGridExpert:
Oh, okay, sorry, I misunderstood what you meant. Changed it

TheGridExpert:
I had to add an extra debug message though, as it doesn't warn you if you write an id correctly but the item doesn't exist. The ResourceLocation parsing shows if it can be parsed at all and doesn't contain invalid characters, while this additional check shows if the item with such an id exists.

TheGridExpert:
Added a few new features regarding vampire books:
- Now it is possible to reroll empty vampire books which will appear if you update from older versions. A new random book is chosen from the tag "vampirism:is_general".
- Not such a big change, but if a vampire book was added from another mod, the name of this mode will be displayed as the mod the book was added from, just like it works with enchanted books or potions.
- Also made a few more book translations.

Cheaterpaul:
I will merge this on the weekend. After that please create separate context specific pull requests.

PR #1491, opened 2025-04-19
Registry Sort

TheGridExpert:
That would be more of a technical pull request, though there are a few new features.

Will start with the fact that ModBlocks and ModItems classes were sorted. By this I mean that I put everything nearly like it is in the creative menu, though with a few differences, for better readability and the speed of finding a specific block. A questionable part, but I've also moved the majority of properties built into block classes to the registry just in case someone decides to make a child class and make them more readable and comparable. For the same reason block flammability, block strippability, and wood types were moved to ModBlocks as a separate method rather than the blocks themselves, thus getting rid of separate classes for wood-related blocks only made for it. Regarding those, dark and cursed spruce blocks now copy the properties of vanilla spruce variants directly rather than just imputing the same values, they are same anyways.

Block classes, as mentioned, also got some changes. From the general ones, I've updated the parameter names of methods to the ones used in Parchment mappings. That's not much, more of a visual thing not impacting on anything. Shapes of some blocks were also changed, for example the altar pillar and the altar of infusion. The reason for this is that the particle number depends on the shape: the bigger the square, the more particles there are generated. That is also the case for coffins, but I just decided not to touch them. Not only that, but the block state of the hunter table, or the variant property to be exact was split into three separate ones. It just allows a simpler block state file while keeping the same point calculation and getting rid of an enum. Item classes were also affected, the biggest ones are the injections. Also tweaked the model a bit to fit better with the new garlic texture as it is not bigger and the old model kinda "drowned" in the desk surface. I also decided to make injections less hardcoded, just in case some other mod decides to add their own one, and made InjectionItem an abstract class while putting the functionality of both garlic and sanguinare injections into their separate item classes. Also made a separate block class for cursed hanging roots so that they, just like their placed placed variant, can only be planted below cursed earth.

Some other changes were also made, mostly involving dyeable items. I've put coffins into the same list, but now I split all of them from Helper to a separate ColorListsUtil class.

Parameters and methods of block and item classes are now also marked notnull by default. That's how it's made in vanilla to remove the need in constant @NotNull before all parameters and makes the code look cleaner.

Tweaked death message ids to ensure that they don't overlap with the other ones, just added the "vampirism" prefix to those which didn't have any. Not only that, but also added a new message for dying after leaving a faction which is "[Player] renounced their nature and perished". Thought it would make more sense than just "died".

Moving onto the actual features - descriptions. Some of the blocks now have descriptions which you can show if you hold shift. They mostly just explain how to use a block, more like a mini wiki. Currently, there are very few of them, only for candle holder blocks and garlic diffusers. The last ones explain the differences between all three versions as some players, from what I've noticed, got confused why the "weak" diffuser has a five chunks radius while the strong one is only one. Their names were moved from descriptions to names of blocks to differ them more easily. (Also thought of renaming them into something else because the term "weak" really sounds confusing as for something the feature of which is the distance, but couldn't come up with something that sounds better).

Not a big feature, but there are now custom splashes for the title screen of the game. Those mostly include vampire-related references. Here's the full list:
- ```Creature of the night!```
- ```Pureblooded```
- ```What a horrible night to have a curse``` - (From Castlevania)
- ```Reject humanity!```
- ```THE WORLD``` - (A quote from Jojo, probably the most popular one. It's basically a vampire-centred franchise, so thought it would be a sin not to include this)
- ```How many breads have you eaten in your life?``` - (Also from Jojo)
- ```Also try Castlevania!``` - (Minecraft has a few splashes like "Also try Terraria!", so decided to add one with Castlevania. Really, try it as well)
- ```What is a man? A miserable little pile of secrets``` - (From Castlevania)
- ```Are you vegan? - Yes, we kill only animals```
- ```Кровосіся``` - (A popular word in the Eastern European community essentially meaning "Bloodsucker", taken from one of the Twilight dubs)
- ```Blood wine tastes the same as I remember... But where are those who share this memory?```
- ```MOTHER```

There's also another rare splash which only appears on Vampirism's birthday date, August 11th, it's ```Happy anniversary, Vampirism!```. Wouldn't mind at all if you added some of your own as well :)

I think I've included everything, hope I haven't missed anything.

But that's not all. There are a few... issues. I've noticed that you cannot open the weapon table menu. At all. The thing is that it basically *must* work, but it doesn't. I tried debugging it, but it looks like the method which is supposed to be called when the player clicks the block just... isn't called. At least for block entity blocks. Not even the container ones or normal blocks. I've tried adding another block with a block entity and it seems to be true: I added a default block with the method which displays the player a message when it's clicked, it worked. Then I changed it to a block entity block and kaboom: it stopped working. What's more, maybe this was in older versions and I just didn't notice because I've started working with block entities quite recently, but when I tried to test a block entity renderer for the same test block displaying the item inside the item was null, though it was not, I checked with /data command. Really weird, not sure what to do with all of that /:

For the same reason I couldn't make diffusers emit particles. I wanted to make them generate smoke when working but for some reason it just didn't work, though it had to. Left the code for particles, just commented out.


src/main/java/de/teamlapen/vampirism/blockentity/diffuser/DiffuserBlockEntity.java
```diff
@@ -195,6 +194,24 @@ public void deactivateEffect(Level level, BlockPos blockPos, BlockState blockSta
 
     }
 
+    public static void clientTick(Level level, BlockPos blockPos, BlockState blockState, DiffuserBlockEntity blockEntity) {
+        /*
```
Cheaterpaul: what's up here?

src/main/java/de/teamlapen/vampirism/blocks/diffuser/GarlicDiffuserBlock.java
```diff
@@ -1,111 +1,57 @@
...
-    private final Type type;
+    private final EnumStrength strength;
+    private final Supplier<Integer> radius;
 
-    public GarlicDiffuserBlock(BlockBehaviour.Properties properties, @NotNull Type type) {
-        this(type, BlockPropertiesExtension.descriptionWithout(properties, "_weak|_improved|_normal").mapColor(MapColor.STONE).strength(40.0F, 1200.0F).sound(SoundType.STONE).noOcclusion());
-    }
-
-    public GarlicDiffuserBlock(@NotNull Type type, @NotNull Properties properties) {
+    public GarlicDiffuserBlock(Properties properties, EnumStrength strength, Supplier<Integer> radius) {
```
Cheaterpaul, line 37: I am not exactly sure if this is relevant as I do not know at what time the codec is used, but if the config option is changed on the server. Then the client does not know about this and behaves differently.

This, of course, is only relevant if the option is changed on runtime and easily fixable by reconnecting. But if possible i'd like to change this.

Maybe have a config codec that serializes the config supplier as id and deserializes it again as config supplier. What do you think?

src/main/java/de/teamlapen/vampirism/blocks/diffuser/package-info.java
```diff
@@ -0,0 +1,7 @@
+@MethodsReturnNonnullByDefault
+@ParametersAreNonnullByDefault
+package de.teamlapen.vampirism.blocks.diffuser;
```
Cheaterpaul, line 3: It seems like JetBrains created a new @NotNullByDefault annotation in the latest release. Even though it is marked experimental, I would just use it instead, as it even covers attributes.

Cheaterpaul (changes requested):
Looks good 🦄.


Seems like we need to use the shift tool tips for more things

src/main/java/de/teamlapen/vampirism/blockentity/diffuser/DiffuserBlockEntity.java
TheGridExpert: As mentioned in the pull request, tried to make diffusers generate smoke particles while they are working, but for some reason it wouldn't define whether the diffuser is working or not. Tried putting it inside of the diffuser block class, but it still refused to work. That's why I left it like this (because the smoke generation code works, it just doesn't understand when it has to run) and asked about it in the pull

src/main/java/de/teamlapen/vampirism/blocks/diffuser/GarlicDiffuserBlock.java
TheGridExpert: Makes sense, will implement it

src/main/java/de/teamlapen/vampirism/blocks/diffuser/package-info.java
TheGridExpert: Ok, nice, didn't know

src/main/java/de/teamlapen/vampirism/blockentity/diffuser/DiffuserBlockEntity.java
Cheaterpaul: sry, didn't read it again.

I think the data is not synced to the client. Only the the diffuser screen gets the progress over the menu from the server.

src/main/java/de/teamlapen/vampirism/blocks/diffuser/GarlicDiffuserBlock.java
TheGridExpert: I've looked, now I'm unsure. Did you mean making something like a map to store keys for configs? That's just the only solution I see as config values do not really have ids on their own we can use to get their values later. Therefore the only way of doing something similar is to make a map where configs would be stored with their keys

src/main/java/de/teamlapen/vampirism/blocks/diffuser/package-info.java
TheGridExpert: Screenshot 2025-04-26 at 08 34 16
I assume that it somehow works separately from the other two annotations. Unless this annotation is in the parent classes, it wouldn't work.

Cheaterpaul: They all work the same. Might not work for you correctly when not using the newest idea version or gradle idea plugin version. Have you tried reimporting the gradle project?

If is doesn't work, keep it as it is know and we can change it later.

src/main/java/de/teamlapen/vampirism/blocks/diffuser/GarlicDiffuserBlock.java
Cheaterpaul: I have taken a look into where it is used. Nowhere. So keep your implementation and add a comment "replace with a config codec if needed in the future".

src/main/java/de/teamlapen/vampirism/blockentity/diffuser/DiffuserBlockEntity.java
Cheaterpaul: The furnace does this using a `lit` block state property `FurnaceBlock#animateTick`. You can add a property to the diffuser.

TheGridExpert: I did try it. But this creates another problem, when I make a state change like it's done in other block entity classes, it just throws all the contents of the diffuser out while keeping the rest of the data for some reason. I didn't manage to get why either

Cheaterpaul: can you commit how you have done it?

TheGridExpert:
It wasn't exactly the same, but tried replicating it to be similar to what I did before. It doesn't change the state (even if I use Block.UPDATE_CLIENT), but throws all the garlic inside out. Removing the setChanged method call doesn't resolve it either.

src/main/java/de/teamlapen/vampirism/blocks/diffuser/package-info.java
TheGridExpert: I've looked it up. Looks like they are incompatible because of the way they work. I assume that the errors will stay unless the class is annotated with those two separate methods, the warnings must be built into the annotation

Cheaterpaul: They should work the same. The way these annotations work are that the tool responsible for compiling (in this case Intellij) needs be be aware on how the annotation should work.

The `@MethodsReturnNotnullByDefault` i think is registered though neo gradle as annotation preprocessor or something and the `@NotNullByDefault` by the idea plugin and Intellij itself.

That way the annotation preprocessor marks the targets as not null. Then the compiler or something just checks if the target is marked (not though the annotations itself, but though the result of the preprocessor).

Here the method is not marked as not null, which seems like the preprocessor didn't do its job or the preprocessor not knowing about the `@NotNullByDefault` annotations behavior, which results in intellij complaining.

But keep it is know. The problem might fix itself in the future, then we can change it if necessary. (Maybe neoforge will adapt it)

Cheaterpaul:
Heh, fixed i in the `VampirismBlockContainer` there was a wrong check (Was wondering why the items drop but the rest stay. A breakpoint in `Container#dropItemStack` did the trick of finding the cause).

TheGridExpert:
Nice. I'll tweak the code to make sure it works and make another commit 👍🏻

(PS: Wait, you've fixed it already, nevermind)

PR #1495, opened 2025-05-06
Vampire Book GUI Changes

TheGridExpert:
This pull request mainly focuses on the vampire book screen, which has been modified, as may be guessed from the title.

Backgrounds are stored in assets and specified using their ids in the book contents file. It includes the information how the book should be displayed, the texture, text color, etc. This way no registry is required and it can be changed using resource packs, same for the book contents files. They're all data-generated.

Regarding those, they no longer store  only the text, but the background id, as mentioned above, and image entires. At first, en_us.json is loaded as default. If the translated file is present, it overrides the fields present there. Images have ids so that you could override the ones with text if needed while not duplicating the rest. So, for example, you need an image which is a fancy title and another one which is a drawing. You don't need to override the drawing, there's nothing to translate, so you just add a new image into the list in the file of another language with an id "1". Though, this way you can also tweak the position of the image on the screen in case it is located somewhere in the middle, so may also be useful even for images with no text.

Here's how the contents file of the "WANTED" book looks like now, for example:
```json
{
  "contents": [
    "..."
  ],
  "background": "vampirism:poster",
  "images": [
    {
      "id": 1,
      "texture": "vampirism:textures/gui/vampire_books/images/wanted.png",
      "page": 0,
      "x": 14,
      "y": 14,
      "width": 149,
      "height": 39
    }
  ]
}
```

There are a few variations of backgrounds, but by default two pages are displayed as if you actually opened a book, or a diary to be more exact:

<p align="center">
Screenshot 2025-05-06 at 00 12 51 2
</p>

Alternatively, as of now, there are two other backgrounds: a letter and a poster. Letters are used for the vampire books which are, well, letters, such as "Dear Martha" or "Blood Infusion Breakthrough". Unlike the diary, it is single-page and only has one sprite. Poster is used only for "WANTED". The difference from the letter is that there's only one paper visible and not a pack of.

<p align="center">
Screenshot 2025-05-06 at 00 21 07
</p>

<p align="center">
Screenshot 2025-05-06 at 00 21 35
</p>

A new vampire book - "Ashes of Past Downs", which is a special one for the advanced hunter Dual-Wielding Grid. It mentions his backstory up to the point when he decides to go into hunters, but not after. Currently thinking about mentioning him in the books of other characters, I have a few im mind, so that the reader could "solve the puzzle", something like that.

Added one translation for "The Journal of a Mad Man", the Ukrainian one, and fixed dashes in the other ones to use the correct ones.


src/api/java/de/teamlapen/vampirism/api/components/IVampireBook.java
```diff
@@ -20,5 +22,11 @@ public interface IVampireBook {
 
     MutableComponent title();
 
+    IBookContents bookContents();
+
     List<MutableComponent> contents();
+
+    IBookBackground background();
+
+    List<IBookContents.IImageEntry> images();
```
Cheaterpaul, line 61: Can you add some documentation to the methods. It would be helpful to directly know what the images are for and what the difference between bookContents and contents is.

src/api/java/de/teamlapen/vampirism/api/general/IBookContents.java
```diff
@@ -0,0 +1,31 @@
+package de.teamlapen.vampirism.api.general;
+
+import net.minecraft.resources.ResourceLocation;
+
+import java.util.List;
+
+public interface IBookContents {
```
Cheaterpaul, line 10: Same here what are the fields for and why does the IBookContents have images and the IVampireBook as well?

src/api/java/de/teamlapen/vampirism/api/general/IBookBackground.java
```diff
@@ -0,0 +1,41 @@
+package de.teamlapen.vampirism.api.general;
+
+import net.minecraft.resources.ResourceLocation;
+import org.jetbrains.annotations.Nullable;
+
+public interface IBookBackground {
```
Cheaterpaul, line 10: Here you do not need to add docs to every method, but please add an interface doc to quickly describe how to use this class

src/main/java/de/teamlapen/vampirism/data/provider/BookBackgroundsProvider.java
```diff
@@ -0,0 +1,22 @@
...
+
+    public BookBackgroundsProvider(PackOutput output) {
+        super(output);
+    }
+
+    @Override
+    protected void registerBackgrounds(BiConsumer<ResourceLocation, BookBackground> output) {
+        output.accept(ModVampireBooks.DIARY_BACKGROUND_ID, ModVampireBooks.DIARY_BACKGROUND);
+        output.accept(ModVampireBooks.LETTER_BACKGROUND_ID, ModVampireBooks.LETTER_BACKGROUND);
+        output.accept(ModVampireBooks.POSTER_BACKGROUND_ID, ModVampireBooks.POSTER_BACKGROUND);
```
Cheaterpaul: If the ModVampireBooks.* are only used to be registered here you can create them here and not store them as field in ModVampireBooks

src/main/java/de/teamlapen/vampirism/misc/BookBackground.java
```diff
@@ -0,0 +1,168 @@
...
+        addProperty(json, "textY", this.textY, 16);
+        addProperty(json, "pageNumberXOffset", this.pageNumberXOffset, 79);
+        addProperty(json, "pageNumberYOffset", this.pageNumberYOffset, 22);
+        addProperty(json, "pageButtonXOffset", this.pageButtonXOffset, 22);
+        addProperty(json, "pageButtonYOffset", this.pageButtonYOffset, 12);
+
+        return json;
+    }
+
+    public static BookBackground decode(JsonObject json) {
```
Cheaterpaul: Why don't you use a Codec for the decoding and encoding?

Cheaterpaul: Ah i see. Too many field :D

We can either add a custom group method. Or even better create subclasses for field that belong together. I would make a class for the page offsets and one for the text position.

What do you think?

src/main/java/de/teamlapen/vampirism/misc/BookContents.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.ArrayList;
+import java.util.List;
+
+import static de.teamlapen.vampirism.util.VampirismGsonHelper.*;
+
+public record BookContents(List<String> contents, ResourceLocation background, List<IImageEntry> images) implements IBookContents {
+
+    public static final BookContents EMPTY = new BookContents(List.of(), null, List.of());
+
+    public static BookContents decode(JsonObject json) {
```
Cheaterpaul: here as Codec as well?

src/api/java/de/teamlapen/vampirism/api/general/IBookContents.java
```diff
@@ -0,0 +1,31 @@
...
+
+        int id();
+
+        ResourceLocation texture();
+
+        int page();
+
+        int x();
+
+        int y();
```
Cheaterpaul: maybe rename to xPos or xOffest? Does that make sense?

Cheaterpaul (changes requested):
Fancy

They look very cool, I especially love that the first and last page look different now.

Looks good, would be good if you add some documentation the to new interface classes, especially when there are multiple methods that are named somewhat the same or rename to make the methods understandable. (no need to add documentation to obvious methods)

Regarding the serialization i would like to use Codec there if possible.

src/api/java/de/teamlapen/vampirism/api/general/IBookContents.java
TheGridExpert: Yes, didn't think of that

src/main/java/de/teamlapen/vampirism/misc/BookBackground.java
TheGridExpert: To be honest, not too sure. At first I used Codec, the same way you suggested (more or less), but it looked... too huge, so I decided to split it into simply parsing everything manually. It all looked like this

Screenshot 2025-05-10 at 02 06 13

Well, if it doesn't impact on the loading delay, then why not. Will get them back in one of the commits

Cheaterpaul: That is totally fine. It is just a record with an interface, so all relevant information you need to understand the class is in the interface.

TheGridExpert:
Made everything asked 👍🏻
Now codecs are used instead and there are also descriptions for all interfaces and some confusing methods in the vampire book builder.

Cheaterpaul:
I moved the book content and background loading part into resourcelistener, so that we do not make I/O on runtime. And I renamed the contents field to translations, to make it clear that it is the text and not the whole content.

And I inlined the background objects. No need to hold a reference of the objects if they are data loaded anyways. We no not need to keen a fallback value, because the items from the datapack can not be removed. Retrieving the object using the key is enough.

Cheaterpaul (approved):

TheGridExpert:
Nice. Didn't use listeners at first because I thought it would load the system more, but yes, that variant looks better, makes sense.

PR #1509, opened 2025-05-27
Renamings and Datagen Changes

TheGridExpert:
This pull request is only technical, mostly involving renaming some classes and cleaning data generation up a bit. Changes are not that essential, but they should make the code more readable.

### ModTiles -> ModBlockEntities
Renamed the block entity registry class to make it clearer and so that it fits into the rest of the code better. I know that block entities used to be called tiles in the past, but now it's quite confusing.

### BESR -> Renderer
All block entity renderers, also known as "BESR", were renamed to "Renderers", for example "AltarInfusionRenderer". "BESR" looks too bulky and they're also called "Renderers" in vanilla, so I assume it would make more sense. Also removed the VampirismBESR class because it only had one method which was never used.

### Effect -> MobEffect
The mob effect classes have been renamed from "(Effect Name)Effect" to "(Effect Name)MobEffect" to match the vanilla ones.

### Mixins
Deleted mixins and accessors that have not been used for a significant amount of time, but did not touch the ones that are in progress, the commented-out ones.

Renamed invoker methods from "invoke_(original method)" to "invoke(Original method)" to avoid the underscore and make them look more like the other methods. The "vampirism$" prefix was also added to some methods that are identical to the original ones.

### Loot Table Injections
Changed the way Vampirism's loot tables are injected into the vanilla ones. Previously it was done manually, just by injecting everything directly. Now they use global loot modifiers provided by NeoForge which is supposed to make the same but with less methods.

### Data Generation Changes
All the datagen classes were given the "Mod" prefix to make them less likely to get confused with the vanilla ones. "RecipesProvider" (the parent class from the mod) was given the name of "VampirismRecipeProvider" as it may be used in addons as well.

The recipe provider was changed. Just split all the builders into a column rather than a line to make it more readable and editable. Some repetitive recipes were made into methods.
Also removed the split into categories of "general", "vampire" and "hunter" as the categorisation was quite abstract and didn't always match, and not all recipes could fit into any of them. Not sure, but maybe adding such category folders but for different recipe types, for example "smelting" and "infusing"?

The loot table provider was split into different classes for the ease of navigation.

### Other
Some long-unused utility classes and methods were removed. Some had been left unused for years.


Cheaterpaul:
I hope you had some fun with all the data generator files :D

Cheaterpaul (approved):

Cheaterpaul:
If you solve the merge conflict, then i will merge it.

TheGridExpert:
Ok. Though, how do I merge the dev changes to this branch? I have seen you do that before, but I'm unsure how to do that as I'm afraid to ruin it if I do it incorrectly.

PR #1514, opened 2025-07-03
Blood-Related Stuff

TheGridExpert:
This pull request is mainly focused on the blood itself and related blocks.

### Blood Fluid
Blood can now be poured onto the ground like water or lava. It is slightly thicker than water in terms of flowing and moving inside, and does not negate fall damage as effectively, while extinguishing fire isn't even possible. When it comes into contact with lava, netherrack is formed, which is especially useful for mod packs with Create or other technical mods. If it touches blue ice, it freezes to create cobbled dark stone.

https://github.com/user-attachments/assets/017b1332-2209-45ec-b627-1981218ced56

### Blood Bottle (and fluid amounts in general)
Blood bottles now store 250 mb of blood instead of 900 as it used to. The reason is that one bucket, which is one block of this fluid, is 1000 mb (or 1 b), meaning that that tiny glass holds nearly as much as the bucket. It made it confusing in modpacks where one bottle usually contains 250 mb or around so. As a compensation, filled blood bottles now stack up to 16, same do blood containers. Each bottle has 5 stages instead of 9, 50 mb each.
The blood container can now contain up to 5000 mb of any fluid that is either the mod's blood or is from other mods and can be converted into it.

Blood bottles now hold 250 ml of blood instead of 900 ml as they used to. This is because one bucket, which is one block of the fluid, is 1,000 mB (or 1 B), meaning that the tiny glass bottle holds nearly as much as one whole block. This made it quite confusing in modpacks, where a bottle usually contains around 250 mb, so I changed it to this valua as well. As compensation, filled blood bottles and blood containers now stack up to 16. Each bottle now has 5 stages 50 mB each and are drunk at once.

Blood containers can now hold up to 5,000 mB of any fluid that is either the mod's blood or from other mods that can be converted into it.

### Blood Grinder and Blood Sieve
Both the blood grinder and the blood sieve have been remade from scratch. The blood grinder now accepts meat and other grindable foods from the top, although these can also be put inside via a hopper. To run, it requires a redstone signal and fabric filters, which can be inserted by clicking or though a hopper on the side of the block. These break eventually, one filter can be used to grind 48 items. When everything is prepared, the grinder starts producing blood, which is poured into the inner tank and then into any block supporting blood below, such as the blood container or the altar of inspiration. Four items can be ground at once every 400 ticks, making it a little faster than before.

The block sieve is no longer required in combination with the grinder, that's what filters are for. It is now solely used for processing blood from other mods. It now looks like a pipe with a filter, pouring blood from the top container to the bottom one. The filters are inserted in the same way and can process 4,800 mB of fluid.

You can now also drink from any fluid tank as a vampire, not only the blood container, but also the blood grinder, the blood sieve and other blocks with fluid tanks from other mods that support output.

https://github.com/user-attachments/assets/05eacca3-6545-4f10-82b4-d87d596ec480

New recipes:
Screenshot 2025-07-03 at 11 16 44
Screenshot 2025-07-03 at 11 16 39
Screenshot 2025-07-03 at 11 16 59
Screenshot 2025-07-03 at 11 17 14

### Other Changes
The authors of vampire books are now mentioned in their content files.

The algorithm for adding splashes to the title screen is now more random, so all of them have an equal chance of appearing. The splash "MOTHER" is now rendered in red. A few new ones have also been added:
- `Blood grinder! Fill it with blood, throw it against the wall!` (A reference to Deltarune Chapter 3, to one of Tenna's phrases, changed to the blood grinder)
- `The Unknown Was Her?` (A direct reference to "U.N. Owen Was Her?" (read like "unknown was her"), which is the boss theme of Flandre from Touhou, who is also a vampire, but it also refers to the unknown authors of vampire books as there are too many of those)
- `Do you know who's spreading the black mist?` (In one of the Touhou games, the main plot was that a vampire princess had spread red mist all over the world so that she could walk out without an umbrella. However, the main protagonist had to find out who was responsible and remove the mist so that ordinary people could breathe again. That sounds way too close to the vampire forest fog from Vampirism, and so the colour of the mist was changed to black, referring to the mod)
- `Hey Mister. I want to hit you heart with my stick` (Refers to the "Hey Mister" meme from Touhou, but now it's about a vampire hunter who wants to kill a vampire with a stake)


src/lib/java/de/teamlapen/lib/lib/blockentity/NetworkedBlockEntity.java
```diff
@@ -0,0 +1,70 @@
...
+        loadMetaData(tag, lookupProvider);
+    }
+
+    @Override
+    public void onDataPacket(Connection net, ClientboundBlockEntityDataPacket pkt, HolderLookup.Provider lookupProvider) {
+        loadMetaData(pkt.getTag(), lookupProvider);
+        setChanged();
+    }
+
+    public void loadMetaData(CompoundTag tag, HolderLookup.Provider lookupProvider) {
```
Cheaterpaul: The class is specifically for using this method? So we could make them abstract?

src/main/java/de/teamlapen/vampirism/VampirismMod.java
```diff
@@ -152,18 +152,14 @@ private void enqueueIMC(final @NotNull InterModEnqueueEvent event) {
     private void registerCapabilities(@NotNull RegisterCapabilitiesEvent event) {
         event.registerItem(Capabilities.FluidHandler.ITEM, (item, b) -> new BloodBottleFluidHandler(item, BloodBottleItem.CAPACITY), ModItems.BLOOD_BOTTLE.get());
         event.registerItem(Capabilities.FluidHandler.ITEM, (item, b) -> new FluidHandlerItemStack(ModDataComponents.BLOOD_CONTAINER, item, BloodContainerBlockEntity.CAPACITY), ModBlocks.BLOOD_CONTAINER.asItem());
-        event.registerBlockEntity(Capabilities.FluidHandler.BLOCK, ModBlockEntities.BLOOD_CONTAINER.get(), (o, side) -> o.getTank());
```
Cheaterpaul, line 155: I would rather keep all capability registrations in one place or at least split by type in ModTiles and ModItems, instead of putting a few in the blocks.

Splitting them up and using EventBusSubscriber makes it harder to find them when you don't exactly know what you are looking for. 

I would generally advise against using EventBusSubscriber on classes that are not dedicated to events, it makes it much harder to find where things happen. Using a dedicated class and then calling a method in a context-specific class is argumentatively better, but that highly depends on the context.

src/main/java/de/teamlapen/vampirism/mixin/client/SplashManagerMixin.java
```diff
@@ -28,45 +30,57 @@ public class SplashManagerMixin {
...
+        vampirism$LOGGER.info("Successfully loaded and added {} Vampirism splashes {} times each. Final list size: {}", vampirism$customSplashes.size(), timesAdded, this.splashes.size());
+    }
+
     @Inject(method = "getSplash", at = @At("HEAD"), cancellable = true)
     private void getCustomSplash(CallbackInfoReturnable<SplashRenderer> cir) {
         Calendar calendar = Calendar.getInstance();
         calendar.setTime(new Date());
         if (calendar.get(Calendar.MONTH) == Calendar.AUGUST && calendar.get(Calendar.DATE) == 11) {
             cir.setReturnValue(new SplashRenderer("Happy anniversary, Vampirism!"));
+        } else if (calendar.get(Calendar.MONTH) == Calendar.NOVEMBER && calendar.get(Calendar.DATE) == 16) {
```
Cheaterpaul: At least add a check if Vampire's Delight is loaded.

src/main/java/de/teamlapen/vampirism/mixin/client/SplashManagerMixin.java
```diff
@@ -28,45 +30,57 @@ public class SplashManagerMixin {
...
+        int timesAdded = (int) Math.ceil((chance * originalSize) / ((1 - chance) * vampirism$customSplashes.size()));
+
+        for (int i = 0; i < timesAdded; i++) {
+            for (String splash : vampirism$customSplashes) {
+                splashes.add(vampirism$RANDOM.nextInt(splashes.size() + 1), splash);
+            }
+        }
+
+        vampirism$LOGGER.info("Successfully loaded and added {} Vampirism splashes {} times each. Final list size: {}", vampirism$customSplashes.size(), timesAdded, this.splashes.size());
+    }
```
Cheaterpaul, line 64: Just a general question, why we do not do this all in the prepare method?

src/api/java/de/teamlapen/vampirism/api/VReference.java
```diff
@@ -19,7 +19,7 @@ public class VReference {
     /**
      * One blood in the players blood stats represents this amount of mB fluid blood
      */
-    public static final int FOOD_TO_FLUID_BLOOD = 100;
+    public static final int FOOD_TO_FLUID_BLOOD = 50;
```
Cheaterpaul: This means that you only need to drink half as much blood?

Cheaterpaul (changes requested):
Nice.

## Grinder

The changes to the grinder and sieve are neatly done, and the interaction feels good. The one though i had, was that we could invert the redstone signal. So that the grinder is on by default unless disabled. This might make it easier for players to use It.

### Hopper & Input

As you noticed as well, the hopper mixin might not be a good idea given some unforeseeable consequences.
My solution to this problem would be to allow hoppers and so on to input the items from all sides. This put the problem with the redstone for the player to solve and not us (The item sucking should stay only on the top direction). Additional with the suggested inverting of the red stone signal, that should be very easy to use for player and make it easier to automate.

To automate the grinder fully, we maybe should allow to put the filter in the hopper as well through hoppers.


## Blood

I never liked the idea of the blood fluid as block, because then you might want to thing about what effects you have when touching it or if you can drink from it and so on. But we can keep it as it is now.
But i would make the fluid slower, due to its high viscosity compared to water and with that reduce the texture speed.

src/main/java/de/teamlapen/vampirism/mixin/client/SplashManagerMixin.java
TheGridExpert: Made it like that just in case some other mod modifies the code too roughly and breaks it. Though that shouldn't occur most of the time, so I will proboably just put all of that into the prepare method, yes.

src/main/java/de/teamlapen/vampirism/mixin/client/SplashManagerMixin.java
TheGridExpert: Will remove it from here in general, better not to refer to an addon in the main mod.

src/main/java/de/teamlapen/vampirism/VampirismMod.java
TheGridExpert: Initially, I did that becase, when looking for a specific block entity, it's easier to find its capabilities right in its class, though that's quite situative. I will avoid using EventBusSubscriber in such cases in future, didn't know.

src/lib/java/de/teamlapen/lib/lib/blockentity/NetworkedBlockEntity.java
TheGridExpert: Agreed.

src/main/java/de/teamlapen/vampirism/mixin/client/SplashManagerMixin.java
Cheaterpaul: you can use the [ModifyReturn](https://github.com/LlamaLad7/MixinExtras/wiki/ModifyReturnValue). That should not care about other mods interferences.

TheGridExpert:
### Grinder

> The one though i had, was that we could invert the redstone signal. So that the grinder is on by default unless disabled. This might make it easier for players to use It.

Perhaps, but that might confuse players. The contraption is powered by redstone, while filters are used as consumables. Making it powered by default may make it look odd in contrast to the redstone blocks from vanilla and other mods. However, we could just add a redstone torch into the recipe and exlain it like that, and that's why it could be disabled if powered. Not too sure about that in general.

The reason hoppers can be disabled with redstone is becasue they are not powered by redstone, but they probably have a mechanism inside, maybe something like a trapdoor, that stops items if activated.

### Hopper & Input

> My solution to this problem would be to allow hoppers and so on to input the items from all sides. This put the problem with the redstone for the player to solve and not us (The item sucking should stay only on the top direction)

Currently, meat is inputted from top and filters from the side, but ok, that may make it easier. The reason why I made that mixin was becase I placed the lever directly on the grinder and never near it. Doing it like that fixes everything and the grinder gets activated even without touching hoppers, but I should probably mention it in the block description for players just in case.

Screenshot 2025-07-06 at 15 55 42

> To automate the grinder fully, we maybe should allow to put the filter in the hopper as well through hoppers.

It's already like that, forgot to capture it in the video.

### Blood

> But i would make the fluid slower, due to its high viscosity compared to water and with that reduce the texture speed.

It's already slower than water. I calculated the values from the real blood density, but can't help but agree, it's not too visible in normal conditions. I should adjust the values around 1.5 or 2 times. Textures are also not a problem, I forgot about those, will do it as well.

TheGridExpert:
I've made the most of the changes, though there are still a few left. Haven't changed anything in the redstone input of the blood grinder yet, except for removing the mixin for the hopper. Both meat and filters can now be inserted from all sides of the grinder via hoppers except for the bottom, a separate MultipleItemHandler class have also been added for such cases. All of the capabilities have been moved to VampirismMod.

I'm not exactly sure about FOOD_TO_FLUID_BLOOD, though. It looks like it's used everywhere, sso changing it just to adjust the blood bottle might not have been a good idea. Would it be better if it was only the MULTIPLIER that was changed in BloodBottleFluidHandler?

Cheaterpaul:
## Blood

The blood looks much better now.

## Grinder

Priviously the players needed to place a sieve, a grinder, a blood container and the hopper. One of the ideas of changing this was to make it easier to understand what is needed. So we removed the sieve block (now being an item), but added a lever. I assume that all player will never turn the red stone signal off, meaning there is a useless lever required to be placed next to the block.

So adding a redstone torch to the receipt sound like a good idea.


## Blood Conversion

The `FOOD_TO_FLUID_BLOOD` is not really used as food to blood converter, from where it is used is more or less a simple x100 constant that was previously only used for grinder/player blood level conversion.

Your idea to reduce the blood in a bottle is already reduced by setting the `IBottleBlood#MAX_VALUE`
 to 5 => 500ml blood. But it might be better to uncouple the blood bottle from the constant and implement the bottle to ml separatelly.

## Blood Bottles

It does not completely work with the stacked blood bottles

https://github.com/user-attachments/assets/3154e13e-f5f2-41c5-a296-93e038686660

I think the blood bottle that is filled should be stacked again if possible, otherwise the inventory might get stacked

TheGridExpert:
### Grinder
Adding a redstone torch into the recipe does indeed make sense in terms of the way redstone works, and the schemes would be simplier, but... it will still make new players confused, as I've mentioned before. It just doesn't feel right that the grinder is on by default while the rest of redstone blocks are off. There are even glowing redstone indicators on top.

Alternatively, we could just remove the redstone indicators and still add the torch into the recipe. This way the block would stop being a "redstone block" and the design would lose an accent colour, but that should be less confusing at least.

### Blood Conversion
> Your idea to reduce the blood in a bottle is already reduced by setting the IBottleBlood#MAX_VALUE
to 5 => 500ml blood

Not 500 mB but 250 mB, that's why it's 5 x 50 and not x 100. Split the variables in the latest commit.

### Blood Bottles
> I think the blood bottle that is filled should be stacked again if possible, otherwise the inventory might get stacked

I'm afraid I've misunderstood a few times for sure, but that's bound to VampirePlayer. I assume the blood bottles could be made into checking the amount of the blood the player needs so that it doesn't occur, but this way there would only be even more stacks.

Cheaterpaul:
> Alternatively, we could just remove the redstone indicators and still add the torch into the recipe. This way the block would stop being a "redstone block" and the design would lose an accent colour, but that should be less confusing at least.

This sounds like a good idea, if thats ok for you.

Cheaterpaul:
> I'm afraid I've misunderstood a few times for sure, but that's bound to VampirePlayer. I assume the blood bottles could be made into checking the amount of the blood the player needs so that it doesn't occur, but this way there would only be even more stacks.

The situation before was that the player dinks from the bottle, but only consumes 50% of the bottle. So the bottle is set to 50% blood.

Now with the stacked bottles, the player drinks from the stack consuming 50% of one bottle, but cant put the now 50% filled bottle back on the stack, because the stack only has filled bottles. SO the now 50% filled bottle is put in a different stack, that is ok.

But now starts the problem, the next time the player drinks from the bottle stack and consuming 25% of one bottle we have now a situation were we have 1 slot with a 50% filled bottle 1 slot with some 100% filled bottles and one stack filled with 75%(100%-25%). But what are we doing with the stack?
1. Remove the excess blood, clear the bottle and just add the empty bottle to the inventory.
2. In a new slot. So every time the player drinks from the stack their inventory shrinks, as a new slot is occupied.
3. We already have a not full bottle, why dont we merge them?
   - Ok, fill one of them and add the rest as new item to the inventory.
   - Ok, now we have a (not full) stack of full blood bottles, one full bottle and an empty / slightly filled bottle.
   - Ok, repeat that, drink from the stack, get a new ~% filled bottle and again and again.
   - Ok, then inventory is filled with a lot of filled blood bottles and the initial stack. This consumes a lot of inventory.
   - When drinking from the stack, can't we merge bottles that we fill back into the stack, keeping the inventory space managed?
   - This was not a probleme before, because there were no stacked bottles

TheGridExpert:
As said, the blood grinder no longer features anything redstone-related in its design. It's now disabled by a redstone signal instead of being powered by it. However, I've decided not to include redstone torches in the recipe to make it simplier a bit.

When it comes to the blood bottles, they are now consumed fully without any leftovers. The capacity is significantly smaller now, so it's less blood to lose. Usually, players don't eat food when they're nearly full, and that's how vanilla food works. I suppose that would be the easiest, although the laziest, solution without messing with the inventory. Added a variable that defines whether the food or any other blood source leaves spare blood just in case, for example, if we have dishes that contain no blood but are consumable by vampires.

TheGridExpert:
Another thing. I didn't notice it when the vampire book reload listener was added, but it skipped an important part which was in the older variant. Only the changed parts are added to the localized files, while backgrounds and non-text images stay in the English variant, that's why images contain ids. This way translations may only need to contain the text without backgrounds and images that are the same.

Also translated Nocturnal to Russian.

Cheaterpaul (approved):

TheGridExpert:
Removed impure blood as promised. Now it's ready for the merge 👍🏻

PR #1521, opened 2025-09-22
Syringes, Advancements, Tweaks

TheGridExpert:
## Syringes

Empty syringes can now be used to take blood from mobs. Each syringe holds approximately 50 mb (one-fifth of a bottle). This can be done manually or with dispensers, therefore allowing automated blood farms. However, mobs cannot be killed with syringes: it will be impossible to take blood from one when it is on the verge of dying.

Blood from syringes can be poured into bottles on a crafting table. This works with both empty and partially filled bottles, but requires no more syringes than the bottle can hold. Blood can also be poured directly into containers from syringes.

Empty syringes stack up to 64, while filled syringes and injections stack up to 16. All injections now return an empty syringe after being used.

The crafting recipe was changed: now it's slightly more expensive in terms of iron, but it no longer requires glass panes. Previously, the recipe produced 16 panes and required 6 glass, which was inconvenient since most of the panes went unused.

Screenshot 2025-09-22 at 19 46 43

### Garlic Injection

The garlic injection can now be applied to any biteable mob. It makes their blood poisonous to vampires for 3 in-game days. Was thinking about making hunter npcs vaccinate villagers and animals to protect that from vampires, might make an interesting detail.

## Advancements

Added three new advancements:
- Patchouli Knowledge
Requires collecting all of vampire books existing in the mod. Might be a challenge, it is. Visible if Vampire Diaries is unlocked. The name is a reference to Patchouli Knowledge, a character from Touhou, who is a librarian who works in a vampire mansion (almost too fitting).

- Baptism Of Blood
Given for using the altar of inspiration. Goes before the Blood Cult one, just for consistency.

- Texas, 1973
The main requirement is consuming any food rich in blood. The parent advancement is the Blood Stream. The title is a reference to The Texas Chain Saw Massacre film, which takes place in 1973.

As Bloody as It Gets now goes after Blood Cult, which goes after the new Baptism of Blood. Rearranged this as one performs the rituals anyway before achieving the highest level.

## Other Changes

- Added description tooltips for the mother trophy, the hunter table, the injection chair, and all the syringes and injections.

- I swapped the active and inactive cursed logs. The ones sucking blood are now "active", while the normal ones are inactive. The cursed spruce trees include the active ones, but they drop the normal variant without silk touch. This should make things easier and more intuitive for players, as not everyone, especially vampires, has spare holy water to apply to all the logs. There have often been cases where players built houses with these, only for bark to grow on them later, making the whole place unsafe. I have noticed the same with some of the Dracula castles/manors suggested for the competition.

- Players in creative mode are no longer required to hold a specific number of candles to fill empty candelabras and chandeliers.


TheGridExpert:
Changed cursed logs once again. Now they're not just swapped, but merged into one block. Instead of a separate block, there is now a blockstate "active", which changes if needed. In case the active variant is needed to be placed, a data component "active" has been added that can be applied to the block item. If the item placed has it applied, the block is active when placed as well. This can only be accessed in creative mode or by chopping active cursed logs with silk touch applied on the tool.

src/api/java/de/teamlapen/vampirism/api/entity/IExtendedCreatureVampirism.java
```diff
@@ -49,11 +59,6 @@ public interface IExtendedCreatureVampirism extends IBiteableEntity {
     @Nullable
     IConvertedCreature<?> makeVampire();
 
-    /**
-     * set if the bite attacker should get poisoned on bite
-     */
-    void setPoisonousBlood(boolean poisonous);
```
Cheaterpaul, line 55: Can you add this back as deprecated for backwards compatibility?

src/main/java/de/teamlapen/vampirism/core/ModBlocks.java
```diff
@@ -146,11 +147,8 @@ public static class WoodTypes {
...
-    public static final DeferredBlock<CursedSpruceBlock> CURSED_SPRUCE_WOOD_CURED = registerWithItem("cursed_spruce_wood_cured", CursedSpruceBlock::new, logProperties(MapColor.COLOR_BLACK, MapColor.CRIMSON_HYPHAE));
-    
-    public static final DeferredBlock<CursedSpruceBlock> CURSED_SPRUCE_LOG = registerWithItem("cursed_spruce_log", props -> new CursedSpruceBlock(props, CURSED_SPRUCE_LOG_CURED), logProperties(MapColor.COLOR_BLACK, MapColor.CRIMSON_HYPHAE));
-    public static final DeferredBlock<CursedSpruceBlock> CURSED_SPRUCE_WOOD = registerWithItem("cursed_spruce_wood", props -> new CursedSpruceBlock(props, CURSED_SPRUCE_WOOD_CURED), logProperties(MapColor.COLOR_BLACK, MapColor.CRIMSON_HYPHAE));
+    public static final DeferredBlock<CursedSpruceBlock> CURSED_SPRUCE_LOG = registerWithItem("cursed_spruce_log", CursedSpruceBlock::new, logProperties(MapColor.COLOR_BLACK, MapColor.CRIMSON_HYPHAE), CursedSpruceItem::new);
+    public static final DeferredBlock<CursedSpruceBlock> CURSED_SPRUCE_WOOD = registerWithItem("cursed_spruce_wood", CursedSpruceBlock::new, logProperties(MapColor.COLOR_BLACK, MapColor.CRIMSON_HYPHAE), CursedSpruceItem::new);
```
Cheaterpaul, line 151: Here we have a slight problem, even though we map the old id to the new id, the blockstate of the old block is not fully compatible. What i mean is that all preiviously "active" blocks now get "deactived", because the id mapping can only map the id not add additional data to the block.

Generally this is a problem, if there is an easy solution (i doubt that), we could use that solution. Otherwise we can ignore it, the next Vampirism version will have breaking changes with the last one, so we might not care.

Cheaterpaul (changes requested):
Looks good.

As you suggested, the hunters running around vaccinating villagers might be a good ides. As you replaced the generic hasPoisonousBlood by a timer, the villagers in a hunter village should be vaccinated most of the times, so we need hunter that to that. A goal might be the solution.

TheGridExpert:
Have added the setPoisonousBlood(boolean) method back as deprecated. However, yeah, I didn't manage to resolve the problem with cursed log compatibility. I tried using the vanilla DataFix, but it didn't work as we not only add the block state, but also switch two blocks, sadly. Though, this DataFix could work for migrating vampire books from older versions without losing their texts. Not sure if it's relevant now, considering that the 1.21.1 version has already broken books if I recall correctly. I don't see too much sense in this.

Regarding the hunter injection goal, will add it... but not here as I'm currently making something for hunters that will affect this aspect, and hence it would be better if this was added after that change is finished.

PR #1544, opened 2026-01-04
Primary Item Retextures

TheGridExpert:
Empty vampire book

— Fixed vampire book rendering. It seems that Mojang has changed the colour requirements for rendered text; it’s now ARGB instead of just RGB. That’s why vampire books were just empty, despite having the correct number of pages. I also renamed `Language` class to `LanguageUtil` to avoid confusion with the vanilla `Language` and added a few more methods for future use (if needed).

— Removed some textures and translations that have been unused since 1.12.

— This one might be a bit questionable, so I’d like to hear your opinion: I decided to move FactionAPI items into vanilla tabs, as the tab itself is quite small. We could later duplicate those into the main tabs of Vampirism and Werewolves as well, just for convenience. It’s not vital for them to exist in only one tab, that's just a question of what's more useful.

— Moved description tooltips into FactionAPI to make them usable in Werewolves and other mods. They are not accessible via the API itself, though, I’m not sure if that’s possible and how to do that, so I’ve left them in the main part of the mod. That’s also why I added FactionAPI to the mod’s dependencies.

— Moved the med chair and the empty syringe to FactionAPI, as those can be used in other mods (e.g. Werewolves previously had an injection item).

— Fixed a small bug where you couldn't place empty syringes and garlic injections into item frames.

— Added item descriptions for the stake and vampire blood. I initially planned to add one for the garlic finder as well, but I ran into issues moving the new rendering I added on my initial branch, since a lot of that changed in 1.21.11. I’ll re-add this once I’m done.

— Updated textures of some items and village totems (since they’re from FactionAPI). Almost all are made by **Zozozrob** (added him to the credits), some are slightly edited by me to better fit the rest. The remodels and retextures of some other blocks are already done, but I want to finish this pull before it’s too late and then move on to those, especially considering that one block which received the biggest change also includes a ton of visual elements, and I need to get on with the rendering changes in 1.21.11, those are confusing so far ;-;

<div style="display: flex; gap: 10px;">
  
  
</div>

Note: the totems with a faction now just have separate core textures.


projects/vampirism/src/main/templates/META-INF/neoforge.mods.toml
```diff
@@ -101,6 +102,13 @@ versionRange = "${minecraft_version_range}"
 ordering = "NONE"
 side = "BOTH"
 
+[[dependencies.${ mod_id }]]
+modId="${factions_mod_id}"
```
Cheaterpaul, line 106: dont forget to add the factions_mod_id to the property replacements in the build.gradle

Cheaterpaul (changes requested):
Nice, looks good. Some nice improvements and texture updates.

Seems like I missed to add factionapi as dependency in the mods.toml.


One thing though. I do not think that the med chair and syringe should be part of the FactionAPI, the med chair is something specific to vampirism and Werewolves only uses the syringe, because it was easier instead adding a own way.

TheGridExpert:
Okay, yes, I see. Will do 👍🏻

Cheaterpaul (approved):

PR #1546, opened 2026-01-08
Improvements to faction villager spawns and curable converted creatures

maxanier:
Rework Villager Spawn events for faction villages.
- Don't create creatures instances that may be discarded anyway
- Vampire villagers may be cured instead of replaced by a new villager
- Better support for future MCA integration

ICurableConvertedCreature implementations must now provide the cured EntityType.
I don't see a clear reason for always passing the EntityType to the "curing" methods. This type should always be decided by the class itself and may not be known by external classes. By removing this argument, it is also possible to externally use these methods. 

<s>I have not yet been able to test this due to several pre-existing bugs:
- The vampire player action menu does not open (client side error is thrown)
- Vanilla golems seem to attack every creature (even cows) regardless of village faction
- The converted villagers don't have a (rendered) overlay</s>

I will try to investigate these issues first.


maxanier:
Fixed issues. Tested. 
Ready for review? @Cheaterpaul do you have any concerns with these changes?

Cheaterpaul (approved):
I see no problems.

PR #1548, opened 2026-01-16
Faction Tag FRIENDLY_TOWARDS_NEUTRAL

maxanier:
New faction tag that only contains hunters (by default).
Maybe this matches the TODO described in FactionPredicate.
Useful for Golem AI.

Sidenote:
The faction tags are "mirrored" all over the place. E.g. the `NOT_NEUTRAL` tag is  defined in the API, mirrored in the faction core code and again in Vampirism.
This has the advantage that is easy to find, but the following disadvantages:
- Usage analysis only reports usage of the specific reference and not the tag
- JavaDoc is not available in the mirrored locations (or it must be duplicated)

Maybe we want to remove this


Cheaterpaul:
I solved it now that there should not be any tag mirroring.

PR #1555, opened 2026-02-07
Food and Furniture

TheGridExpert:
### Throne (Royal Chair?)

The sit utility has been updated with several new features. It's now possible to lock the player's body rotation and add custom stand-up position calculations (previously, players would simply return to their original sitting location, which looked odd). The throne now utilises this functionality and features a brand-new model created by zozozrob. However, being honest, it has always resembled more of a "royal chair" than a throne, at least for me, and that is how people have been using it since it was added. I have never spotted anyone putting it in a throne room. Perhaps it should be renamed accordingly?

Screenshot 2026-02-05 at 23 55 42 2

### ~~Interrogation~~ Injection Chair

The injection table has also been redesigned. I made it out of wood with leather restraints to better fit the mod's era, and the recipe has been adjusted accordingly. It's now vertical rather than horizontal, so I had to modify the hunter trainer houses as well, those now generate properly.

A new feature allows you to actually use the injection by clicking into the air while sitting on the chair. It's a small addition, but it makes sense given that the chair is meant to be sat on during use. Maybe removing the option to click the block itself if this works then? It might be a bit more logical from the gameplay perspective.

Both chairs are now waterloggable by the way.

Screenshot 2026-02-05 at 23 56 52 2

### Coffins

Coffins have received new models created by zozozrob. The stand-up position is now more appropriate when the coffin is in the horizontal position, though it behaves slightly oddly in the vertical orientation, I'll need to adjust this. It is not possible to test this now though. While coffins generally worked on 1.21.4, they no longer do on 1.21.11 because, apparantly, Mojang changed the bed or world rules. When I first tested them on 1.21.11 I neded up exploding them in the overworld. And, a cherry on top, players can't physically enter the coffin at all — it simply denies entry at all cost. I haven't resolved this yet, but I'll attempt to fix it while you're reviewing the pull request.

Screenshot 2026-02-06 at 00 00 15

### Bat Cage

The bat cage now serves a purpose beyond just decoration. By default, it's empty, but you can catch a bat by clicking on one. This is a bit tricky, but you'll eventually end up with the bat in the cage. You can release it by shift-clicking the ground. It also displays a tooltip with the bat's name in case someone wants to distinguish between different bats. The model was created by zozozrob, though we're still unsure how to properly design the top of the cage. Just a cube looks uninteresting, we wanted something more interesting.

Screenshot 2026-02-06 at 19 18 35 2

### Vampire Forest

The cobbled dark stone has also obtained a new texture. The previous one was too sharp in my opinion. This new texture was created by zozozrob.

Screenshot 2026-02-06 at 00 04 08
(The normal dark stone is just for comparison, it remains the same)<p>

The vampire forest blocks have also been updated. All blocks on the pictures got new textures, including dark spruce, cursed earth, vampire orchid, cursed roots, etc. All by the same artist. However, the bark remains unchanged. I'll come back to it eventually, though I'm currently not sure how to approach it.

Screenshot 2026-02-06 at 00 08 46

Severak adjustments have been made to the foliage. A few parameter adjustments, including the grass colour to blend seamlessly with the blocks. I also changed the falling leaves to needles instead. I honestly hate it how vanilla Minecraft made spruce trees drop regular leaves, that is just laziness.

Screenshot 2026-02-06 at 00 1

## Food

Vampire food is now stored in a single data component that includes food properties, the faction tag, and food behavior. The food behavior is a registry for cases where the faction uses an alternative food bar, such as the vampire blood bar. This is essentially a function that applies food properties to the player or other mobs. All of this is integrated into FactionAPI, meaning that it will be very compatible with everything.

I've also made garlic bread slightly more tasty to hunters specifically. It's not a significant difference, but it's a neat bonus.

### Garlic Finder

Have also revised how the garlic finder reveals garlic diffuser positions. They're now highlighted within a 40-block radius, made sure to implement a cache system to prevent lag, as this practically just scans blocks around the player which might cause some. The highlighting disappears when the player is within a few blocks. The highlight color can be changed in the configs. This should make the item generally more useful.

Screenshot 2026-02-06 at 00 02 04

### Translations

Added a few new Ukrainian vampire blook translations and adjusted puncuation in the existing ones.


projects/faction/src/main/java/de/teamlapen/faction/common/world/items/consume/FactionFoodEntry.java
```diff
@@ -0,0 +1,31 @@
...
+import de.teamlapen.faction.api.FactionRegistries;
+import de.teamlapen.faction.api.factions.IFaction;
+import de.teamlapen.faction.common.core.FactionFoodBehaviours;
+import net.minecraft.network.RegistryFriendlyByteBuf;
+import net.minecraft.network.codec.StreamCodec;
+import net.minecraft.resources.Identifier;
+import net.minecraft.tags.TagKey;
+import net.minecraft.world.food.FoodProperties;
+
+public record FactionFoodEntry(TagKey<IFaction<?>> faction, FoodProperties foodProperties, Identifier behaviour) {
```
Cheaterpaul: Use a Holder of the FactionFoodBehavior and FactionRegistries.FOOD_BEHAVIOUR.get().byNameCodec() as codec

Cheaterpaul:
Very nice. Cool solution for the garlic finder. Nice updated textures with matching style.

PR #1557, opened 2026-02-16
Change the Altar of Infusion

TheGridExpert:
I'm currently trying to move the altar of infusion changes I did on the other branch on 1.21.11, but having problems with the rendering. For some reason, anything that is client-sided, including not only the block entity renderer, but also sounds and particles, are just not there, despite the code looking fine.

The altar model was changed, but the red thing on top was changed into a blood sphere. It is supposed to levitate above the block and rise up when the ritual is running. The server-side interaction is fine though.


Cheaterpaul:
The sound part is a little quirky, minecraft simply uses sound with only the mono channel. ffmpeg can convert it

> ffmpeg -i choir_short.ogg -ac 1 choir_short.tmp.ogg 

And about the model rendering, i would suggest not using the posestack to rotate the whole model, but to give the model's render state the rotation and the rotate the model in setupAnim

TheGridExpert:
It turned out the sounds had a few other problems within the code, mostly regarding how they were played. Managed to fix them and also adjusted the sound files themselves to make those sound better without any clicking. On top of that, I added another sound (beam_enter_player) that plays when the ritual beam enters the player. It is just an edited wind sound, as I don't have much experience with audio, but it seems to fit.

I also had to change the rendering relatively significantly. There was a problem where, even after your changes, not everything was rendering correctly. I fixed the client-server synchronization, improved some visuals, and made the sphere glow properly (as it's supposed to). The blood sphere itself is essentially the catalyst of the ritual: it rises to the point where the beams connect and then condenses them into the player.

Additionally, I’ve changed the altar pillars. These now have new models by zozozrob and slightly different placement conditions (previously, the tip would levitate even if not placed on a block, adjusted such issues). However, there is one new feature that could be interesting for builders or players on servers: you can now click on a pillar block with a slimeball to "glue" it. This makes the filler block unremovable. This is useful for builds (like stone temples where these are used as pillars) so you don't accidentally remove a block, and it prevents people from stealing gold from altars on servers (like Vampirism.co, it's a constant thing). The glue can be removed with an axe, the mechanic is similar to waxing in vanilla, but with slime.

PR #1560, opened 2026-03-01
Potion Table

TheGridExpert:
Another block refinement, now it's the potion table. The usage of the block has remained the same, but the design has changed. Now it's not exactly a potion table, but a distiller. The idea behind it is that hunters distill the vampire blood mixed with reagents like redstone or glowstone, then add the obtained substance into potions. Potions unique to the potion table are made by the same logic, but by distilling other ingredients. As a consequence, the block has been renamed into "Vapor Still", because it's not exactly a table anymore, and this would make it more recognisable. The idea of making it into a distiller and its features are by Random, the model is made by me and Zozozrob, the code and UI is made by me. It's actually quite old, we came up with this long time ago, but this has remained frozen for a year or around so.

The usage has not changed, as said before, same potions and recipes. What changes is the menu. It now reflects the new block model and has proper empty slot icons. The fire icon has the same function as bubbles in the vanilla brewing stand. However, I've adjusted the hopper container handling, so it is more possible to automate this now, and added a tooltip. It now also has copper sounds.

Decided not to take screenshots, here's a demonstration video with everything:

https://github.com/user-attachments/assets/6b38c3ff-3b89-4fd4-af5a-a50a02aa9676


projects/vampirism/src/integrations/jei/java/de/teamlapen/vampirism/common/integration/jei/VampirismJEIPlugin.java
```diff
@@ -82,7 +82,7 @@ public void registerCategories(@NotNull IRecipeCategoryRegistration registration
     public void registerGuiHandlers(@NotNull IGuiHandlerRegistration registration) {
         registration.addRecipeClickArea(AlchemicalCauldronScreen.class, 80, 34, 20, 15, ALCHEMICAL_CAULDRON);
         registration.addRecipeClickArea(WeaponTableScreen.class, 114, 46, 20, 15, WEAPON_TABLE);
-        registration.addRecipeClickArea(PotionTableScreen.class, 145, 17, 9, 28, POTION);
+        registration.addRecipeClickArea(VaporStillScreen.class, 145, 17, 9, 28, POTION);
```
Cheaterpaul: Can you adjust the hover position to match the new location

projects/vampirism/src/integrations/jei/java/de/teamlapen/vampirism/common/integration/jei/categories/PotionTableRecipeCategory.java
```diff
@@ -67,7 +67,7 @@ public void draw(@NotNull JEIPotionMix recipe, @NotNull IRecipeSlotsView recipeS
         this.background.draw(graphics);
         graphics.pose().pushMatrix();
         this.blazeHeat.draw(graphics, 1, 35);
-        this.bubbles.draw(graphics, 3, 4);
+        //this.bubbles.draw(graphics, 3, 4);
         this.arrow.draw(graphics, 80, 10);
```
Cheaterpaul: the new texture is used here but the layout does not match background and is a little broken

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/util/MigrationData.java
```diff
@@ -98,7 +98,7 @@ private static void fixItems(@NotNull Mapping mapping) {
     }
 
     private static void fixBlocks(@NotNull Mapping mapping) {
-        mapping.remap("vampirism:blood_potion_table", "vampirism:potion_table");
+        mapping.remap("vampirism:blood_potion_table", "vampirism:vapor_still");
```
Cheaterpaul, line 101: can you add a new mapping from vampirism:potion_table to vapor_still

projects/vampirism/src/integrations/jei/java/de/teamlapen/vampirism/common/integration/jei/categories/PotionTableRecipeCategory.java
Cheaterpaul: image

TheGridExpert:
I think I forgot to adjust the JEI integration after I was done with the GUI... Will do it 👍

Cheaterpaul:
Other than that, it look way more fancy :rocket:

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/util/MigrationData.java
TheGridExpert: Already there on the bottom of that method

TheGridExpert:
Done! Took... longer than I initially expected, as I've rearranged the GUI layout for 10 times at this point, but now it's pretty similar to the brewing stand JEI gui. The skills are listed in a separate tooltip to save space and are marked green if they're unlocked and red if not.

Screenshot 2026-03-14 at 20 56 42

Cheaterpaul:
Btw. Where did you find out how the jei plugins need to be enabled?

TheGridExpert:
> Btw. Where did you find out how the jei plugins need to be enabled?

It was actually quite random. I got into a dead end trying to make JEI work, so I gave an LLM some context and asked if it had any ideas. It suggested that some mods use services nowadays, so does JEI maybe, if @JeiPlugin on top of the class no longer work, so I should look into this. I found on GitHub 4 repositories, or around so, in total via search, that used a service file for JEI. I tried making the same and it worked somehow. I assume JEI forgot to update their official docs, maybe they mentioned it somewhere, but I couldn't find any info about that on offical resources.

PR #1581, opened 2026-04-17
Syringe Improvements and Serum Injections

TheGridExpert:
## Syringes

Syringes are no longer as primitive in terms of use as just clicking on a mob to collect blood, now they require holding the mouse down. The target gets a 50% slowdown. I am not sure since when this is a thing, but there's a now a method that handles entity interactions right inside the Item class. Haven't seen it before, it must have been added recently.

### HUD

As it can be seen in this video, I've also added a new crosshair that appears while using the syringe. Identical to vampire fangs in terms of functionality, but it looks like a blood drop. Vampire fangs appearing would be weird for non-vampires, hope this looks fitting enough. The fangs themselves were also retextured, they're also insignificantly larger now, but it should make them demonstrate the blood level a bit clearer now.

https://github.com/user-attachments/assets/b5091750-e23e-484d-b437-cfe14f8d7a09

## AppleSkin FactionAPI Integration

I've added an integration for AppleSkin built into FactionAPI. I already implemented this in Vampire's Delight, but if faction food is in FactionAPI, it'd better to add it there. It's not only an integration but a good way to debug food that has faction-specific food properties.

Also tried adding an integration for Vampirism to turn hunger points into blood drops... but then realised that a bunch of stuff in AppleSkin is hardcoded, so it's pretty problematic. Have put it aside for now, but I might just make a pull request for AppleSkin to broaden the access for external mods a bit.

The gradle dependency is also inside Vampirism's code, but, despite it not working, it is required, because the instance doesn't load otherwise if only FactionAPI has access to AppleSkin.

## Serum Injections

This has been on my mind for quite some time, discussed this with Random, but I've got my hands on it only now. Essentially, serum intections are potions that can be used instantly by left-clicking. They stack up to 4, last half as long, go on short cooldown after using a single shot. One unobvious benefit of them is that they cause less lags than splash potions according to Bitul. Serums can be crafted by combining any potion with a syringe on a crafting table. I am not sure if they should be restricted for hunters, so for now they're general. Will change it according to your opinion.

The serum textures are work-in-progress, will be changed.


supermike1999:
Like the design change of the icons, though they're a bit big.

Ede1vveiss:
I also like the new design, I also think it could be interesting if the darker gray would be more translucent while keeping the opaque white edge. The red would stay opaque.

TheGridExpert:
Have changed the icons to be smaller and added 70% transparency to the dark gray part of the backgrounds. What do you think?

https://github.com/user-attachments/assets/b78250f2-e195-488d-964d-d089f39ac877

Also added changes suggested by Bitul and Mike. Now syringes stack up to 2 (was 4) and have 60% of the normal potion duration (was 50%). Added a new skill, Bivalent Injections, that can be unlocked instead of Concentrated and Durable Brewing in the hunter skill tree. It allows combining two potions into one syringe (excluding blocked ones or duplicates). Bitul suggested blocking certain potions from being used in serums, so I added a blockedSerumPotions config in the balance file. Currently, it blocks all variants of instant health, resistance, and turtle master. All injections are hunter-only and can only be crafted at the weapon table, no lava required. That's mainly because it's impossible to get the instance of the server player in a vanilla crafting recipe, which is required to check the skill, so I had to move it there. However, as a consequence, I've rewritten some of the weapon table recipe code so that it's possible to write custom recipes in a similar way to the crafting table.

Screenshot 2026-04-21 at 18 26 40
(The icon is supposed to be two serum injections tied with a ribbon)

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/config/BalanceConfig.java
```diff
@@ -740,6 +746,19 @@ public BalanceConfig(BalanceBuilder builder) {
         itApplicableOilSwordReverse = builder
                 .comment(String.format("When enabled, the '%s' item tag acts as a whitelist instead of a blacklist for sword oil application.", ModItemTags.APPLICABLE_OIL_SWORD.location()))
                 .define("applicableOilSwordReverse", false);
+        itBlockedSerumPotions = builder
```
Cheaterpaul: I think that is something better suited for a Tag, is it not?

Cheaterpaul (commented):
Nice improvement.
Yes, the overlay is much better this way.

I am not 100% behind the serum injection, but when mike has no objections then it is totally fine.

I must think about the appleskin integration. I am not convinced yet, that it is worth the effort maintaining it here (faction + vampirism). Integrations in vampirism directly were always meant to make vampirism work better, while integration in the integrations mod where meant to make other mods work better with vampirism. I'll need to think about it.

TheGridExpert:
The thing with AppleSkin is that it's a useful mod for debugging and survival, but it's not really possible to make it a part of the integrations mod. Now that FactionAPI is separate, we would need another mod for FactionAPI specifically if we want an integration, so it's "cheaper" to just put it into the base mod.

For the vampirism part of the integration the situation is a bit different. Basically, it's not yet done and it's purely cosmetic, the real values are in the FactionAPI part. However, I didn't manage to add the dependency to FactionAPI without adding it to Vampirism, Gradle panics. So we have th dependency anyway. What I could do is making a pull request for AppleSkin to introduce an event or a sub-event for textures (or just unhardcode certain things). This way it would be much easier to maintain and it would break with new versions less. But that's only if they agree.

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/config/BalanceConfig.java
TheGridExpert: Agreed, will turn it into a tag. The initial idea was to allow easier edits by server admins. Knowing Zin, he would like to configure such stuff, but, yes, this is better as a tag for sure.

Cheaterpaul:
Ok so I though about the apple skin integration.
First some explanation. The problem with such integration (same to the jei/guide) is that the mod must be available when we want to publish a version, or we need to push a new release just to enable the integration when it is available. This is typically not a problem on established mc versions, but jei was known to take time to port to new versions. Additionally handling more non neoforge dependencies always slowed down gradle. But this is now way better in newer neogradle/moddevgradle. Additionally more external mods bloat up the dev environment.  That is generally what makes me not like 3d party integrations in the main mod dev environment.

That being said, with the current gradle setup i did, it is much easier to handle integrations and i am ok with adding a appleskin integration to factionapi for now. (subject to change if there is potential for more integrations, but lets first publish the new versions).

But two things that need to be done for me to allow the integration.
1. remove the integration in vampirism for now (works for me)
2. do not use mixins to change the functionality in appleskin. NEVER do that. That just creates more problems. (unmaintained mods are an exception of cause). Please make this work without mixins

There is the FoodValuesEvent, maybe you can create a Pull Request to appleskin to make the consumable editable as well. And the Food check; maybe we can register an empty food properties for our faction food.

TheGridExpert:
Makes sense. I removed the integration from Vampirism and used FoodValuesEvent in FactionAPI. The initial problem was that the sourceSets weren't considered in build.gradle, which is why events didn't work inside integration modules – now they do. The only remaining mixin is the isFood method override, I'll make a PR for it, though probably for the 26.1 version since we're moving to it anyway, so it makes more sense to target that one.

TheGridExpert:
PR: https://github.com/squeek502/AppleSkin/pull/430

Cheaterpaul:
I talked a little bit to mike about the serum injections, while he is not a fan of them, i think with some changes we can them better fit into the mod.

## What wrong with them

One big argument is that with serum injections there is not really a point for other potions, besides splash potions to hit multiple players. This would make hunter potions pointless to a large degree, as they are only used as additional ingredient for serum injections. And I agree here totally it would be better if the serum injections would stand out in a different way than just be a "better" type of potion.

They have basically no weaknesses. Normal potions take time to consume and splash potions can miss or hit your foes. Yes they have cooldown, but if you just one potions there is none.

Making them stackable simply increases you total potions capacity x2 while taking nothing away.

Adding two potions into the serum injection even further increases their power without taking any drawbacks.

This mainly makes all other potions rather pointless.

## Suggestions

Instead I would suggest to modify the serum injections to make them a valid alternative with specific usages instead of a evolution of normal/hunter potions.

- do not allow hunter potions for the serum injections only normal ones. (This could additionally make other factions be able to use them.)
- set the max stack size to 4, but reduce the duration of the effect to 25%
  - So basically split a potion into 4 injections
  - the recipe for short potions may be just 1 -> 2. while for longer it may be 1 -> 4.
- remove the bivalent injection, it is relatively strong and hard to balance
- add some kind of short debuff to give it some weakness. The idea here is that you inject yourself some stuff into your veins. That must have an effect.
  - my idea here was to reduce max health for the duration / half the duration
  - mike suggested some weakness/mining fatigue or nausea

This would overall make a nice addition, without replacing the functionality of the hunter potions, and create new ways of using potions.

## Usecases

With the hunter potions and serum injections can assume that there are ways player can play:

1. Players that are rather dedicated to fights that might take some time. They prepare their hunter potions for the best effects. Use a potion and have a long effect
2. Players that do not want to spend time getting the resources for the hunter potions have at least a ways for some benefit with the serums injections
3. Players that maybe want shorter skirmishes and do not need the maximum effects. They might use serum injections to use less resources and have more that can be consumed fast.
4. Other faction players can get serum injections from hunter player to have something better than just normal potions 



What do you think here @TheGridExpert is that something you can get behind with. Do you have some suggestions or concerns?

TheGridExpert:
I can agree with the most of things, will implement those (though sad for the bivalent injections). However, I'm not sure about the use aftereffect. Reducing the maximum amount of health sounds quite cruel. If I were to do that, I'd rather add some sort of Exposed/Vulnerable effect that amplifies the damage you take while it's active. For example, for 5-8 seconds after using the injection.

Cheaterpaul:
> I'd rather add some sort of Exposed/Vulnerable effect

That sound good as well. It shouldn't be to strong but have an effect. But we can workout details later when testing the whole thing.

TheGridExpert:
> the recipe for short potions may be just 1 -> 2. while for longer it may be 1 -> 4.

I'm a little concerned about this part. I'm afraid it might be unintuitive for some whether a potion requires 4 syringes or 2 in the crafting grid. Maybe just making all split into 2 instead of 4, this would be less messy?

Cheaterpaul:
That was just a though I had, given that some effects might be to short then. But aside from regeneration it seems that the shortest potion time is 1:30. And that is fine to split in 4.

TheGridExpert:
Okay, all done now. I've removed the mixin (that will be readded as an event when/in case the PR is merged) and changed the syringes.

Now one potion is split into 4 serums and you get Exposed effect when you use one. Added a efExposedPerLevelMultiplier config to regulate how much more damage the player takes, 25% per level by default that last 5 seconds after using the serum. And bivalent injections are gone...

However, this, honestly, needs testing, but I'd rather not make those serums abstract from hunters, after all, that's their specialisation. Maybe there could be a skill that allows to mix hunter potions into serums, which would be positioned in the same place where bivalent injections used to be?

maxanier:
It may be a bit confusing that there is a "Empty Syringe" and a "Blood Syringe", but when filling the "Empty Syringe" it becomes an "... Injection"

TheGridExpert:
Yeah, it might be... but changing to a single postfix would be even more confusing. An injection is something that you inject into yourself, so an empty syringe is not an injection on its own. A blood syringe is also not an injection: quite the opposite, actually, you use it to take the blood out of the target, not into it.

Both garlic and sanguinare injections are used on the injection chair, so this is pretty much a hint to the player how the item is used. The empty and the blood syringes do not interact with the chair in any way, but injections do.

maxanier:
Ok, makes sense.
Would you rather use "Syringes" or "Injections" as an umbrella term for a guidebook entry that describes the injection process and syringe crafting process?

TheGridExpert:
"Injections", I think.

PR #1589, opened 2026-04-28
Update to GuideBook

Piklach:
Changed some directory addresses for recipes, filled in missing pages from missing data, updated some of en_us to make clearer/remove added spacing and such. Updated Hunter Levelling requirements chart to match the missing Pure Blood 1's that are required ingame but not previously in guidebook


maxanier:
Perfect, thank you!

PR #1592, opened 2026-05-02
Fix duplication glitch with Sable (Create : Aeronautics)

Daudeuf:
Implement Clearable interface to PedestalBlockEntity to made it work with Sable


Daudeuf:
Do you know when there will be a new update to the mod so that the changes will be available?

Cheaterpaul:
Today or tomorrow.

PR #1607, opened 2026-05-16
add integrations loader

Cheaterpaul:
Adds a `Integration` annotation that can be used to identify integration to other mods.

It works similar to `EventBusSubscriber` with additional checks.

It checks if the mods specified in the Integration attribute are loaded and on the correct dist. If the checks succeed, the static method annotated with SubscribeEvent are registered to the game bus or the owning mods event bus.


Cheaterpaul:
Expand this to add accepted version range to the modids.

Move Balance Config configuration into an event. Then we can use this for vampirism integrations as well.

PR #1610, opened 2026-05-25
Block Remodels and Fixes

TheGridExpert:
### 1.
Fixed the issues regarding the altar of infusion and inspiration.
[ALL REQUIRE ADDITIONAL TESTING]

The Altar of Infusion:
- The ritual is paused if the player leaves the game, cancelled if the player teleports to another dimension or walks 30 blocks away;
- The block cannot be broken by other players if the player is in the game and the ritual is running;
- The slowdown effect is removed in any incorrect outcome of using the altar, the player won't get slowed down forever.

The Altar of Inspiration:
- Organised the code;
- Fixed a bug where you couldn't fill the altar by any means.

### 2.

I've noticed that the level-up sound effect is played automatically when the player changes their level. It is... quite long, so it is very annoying if you level up fast or just use commands. Moreover, the altar of infusion had its own sound — Choir Short, so there were two different sounds playing at once.

I decided to leave only one of them, so I chose Choir Short as it was shorter and more fitting (at least in my opinion; swap back to the previous long level up sound if you find it better). Also added a link for Choir Short in the sound license file and the readme.

The sound is handled in FactionSounds.playLevelUpSoundServer() and called in the level-uppers themselves (altars and hunter npcs), so it's synced. However, for the hunter trainer I had to remove the note block sound effect as the two were layering.

### 3.

Added ColorConfigValue for registering color configs. For now, it's only for garlicFinderAuraColor, but it will be useful in the future for sure as it's in FactionAPI and it makes the code less messy. The colors are refreshed automatically with an event subscription inside the main config class of the mod that uses this value.

### 4.

There is now a separate key bind for item descriptions. Left Shift by default, but can be configured to something else because you usually move items with shift held, so it might be annoying.

### 5.

The fluid rendering inside blocks like the blood container and the blood grinder has been fixed.

### 6.

The bat cage now correctly renders the bat texture, fixed the location.

### 7.

The fireplace block has been retired.

## Block Visuals

A few updates made by me and zozozrob:
- The tombstones have been remodeled. They're also waterloggable now.
- The blood pedestal now has a new model and particles.
- The blood-infused iron blocks now have updated textures.
- Chandeliers got a less bagel-like model.

Screenshot 2026-05-25 at 22 09 23

Also... the mother trophy now *slowly rotates to look at you like a ghost...*


Cheaterpaul:
I honestly do not like the new models/texture.

The tombstones, the altar and the altar of cleansing look too flat and simple in comparison. They might need an updated model, but I still prefer them over the updated models.

The blood iron blocks look fine and the blood tank looks acceptable.
The chandelier though, I like the current one better, they look more like a chandelier to me. While there look more like a basket.

projects/faction/src/main/java/de/teamlapen/faction/common/factions/FactionPlayerHandler.java
```diff
@@ -370,8 +369,7 @@ public boolean setFaction(LevelingChange param) {
             });
         }
 
-        if (this.player instanceof ServerPlayer serverPlayer) {
-            serverPlayer.connection.send(new ClientboundPlaySoundEventPacket(FactionSounds.LEVEL_UP));
```
Cheaterpaul, line 374: I see absolutely no problem in playing the level up sound. The 4 seconds it takes is not relevant. How should a player get multiple level ups in 4 seconds? Having a longer sound makes the level up a little bit more satisfying.

Yes, better to use the same sound everywhere, but i would prefer this one.

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/world/blockentity/AltarInfusionBlockEntity.java
```diff
@@ -230,6 +235,24 @@ public static void tick(Level level, BlockPos pos, BlockState state, AltarInfusi
...
+                }
+                if (onlinePlayer.level() != level || onlinePlayer.isDeadOrDying() || onlinePlayer.distanceToSqr(blockEntity.worldPosition.getX() + 0.5, blockEntity.worldPosition.getY() + 0.5, blockEntity.worldPosition.getZ() + 0.5) > MAX_PLAYER_DISTANCE_SQ) {
+                    blockEntity.player = onlinePlayer;
+                    blockEntity.endRitual();
+                    return;
+                }
+                blockEntity.player = onlinePlayer;
+                blockEntity.ensureSlowdownModifier(onlinePlayer);
+            }
+
```
Cheaterpaul, line 244: If the player leaves, the ritual should be aborted, as well.
No point having a stuck altar when the player is offline for a few hours.

TheGridExpert:
Sorry... I will change what I can.
(By "them" in "but I still prefer them over the updated models" you meant "the previous ones", correct?)

For now, I can just rollback the tombstones and the altar of cleansing to their previous models (but with the same code changes like waterloggability and such).

The chandelier is the questionable part though... I agree with the basket thing, but the initial reason why I changed those was that I've heard quite some complaints related to the previous models that... also called them baskets.

As Spade said in the chat, maybe making a poll about these models? At least the chandelier ones, maybe the gravestones as well (the problem with gravestones is that it's essentially just a brick... I'm not sure what can be changed about it really).

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/world/blockentity/AltarInfusionBlockEntity.java
```diff
@@ -318,21 +331,43 @@ private void applyPostRitualEffects() {
...
+    private @Nullable ServerPlayer resolveOnlinePlayer() {
+        if (this.playerUUID == null || this.level == null || this.level.getServer() == null) return null;
+        return this.level.getServer().getPlayerList().getPlayer(this.playerUUID);
+    }
+
+    // Re-applies the slowdown each server tick. ServerEventHandler.onPlayerLoggedIn strips it unconditionally to clean up stranded modifiers; this restores it if the ritual is still active.
+    private void ensureSlowdownModifier(Player target) {
+        AttributeInstance speed = target.getAttribute(Attributes.MOVEMENT_SPEED);
+        if (speed != null && !speed.hasModifier(ID_MOVEMENT_SLOWDOWN)) {
+            speed.addPermanentModifier(new AttributeModifier(ID_MOVEMENT_SLOWDOWN, -1, AttributeModifier.Operation.ADD_MULTIPLIED_TOTAL));
```
Cheaterpaul: To resolve the issues with the modifier, is to use a `AttrbuteInstance#addTransientModifier` which is the one you should be by default and which is not saved. The addPermanentModifier s for permanent stuff.

Cheaterpaul (changes requested):

Cheaterpaul:
> (By "them" in "but I still prefer them over the updated models" you meant "the previous ones", correct?)

The ones that are on the dev branch at the moment

> The chandelier is the questionable part though

Yes we can make a poll

I feel the current gravestones have more of a character.

projects/faction/src/main/java/de/teamlapen/faction/client/config/values/ColorConfigValue.java
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * If you want to use this value type in your mod, make sure to subscribe {@link ColorConfigValue#subscribe(IEventBus)}
+ * to your event bus; inside {@link Services#registerModBus(IEventBus)} if you're using {@link Services}.
+ */
+public class ColorConfigValue {
+
+    private static final Pattern HEX_PATTERN = Pattern.compile("^#?[0-9a-fA-F]{6}$");
+
+    private static final ThreadLocal<@Nullable List<ColorConfigValue>> ACTIVE_COLLECTION = new ThreadLocal<>();
+    private static final Map<IConfigSpec, List<ColorConfigValue>> BY_SPEC = new IdentityHashMap<>();
```
Cheaterpaul: Use a concurrent map

projects/faction/src/main/java/de/teamlapen/faction/client/config/values/ColorConfigValue.java
```diff
@@ -0,0 +1,100 @@
...
+        List<ColorConfigValue> list = ACTIVE_COLLECTION.get();
+        ACTIVE_COLLECTION.remove();
+        if (list != null && !list.isEmpty()) {
+            BY_SPEC.put(result.getRight(), list);
+        }
+        return result;
+    }
+
+    public static void subscribe(IEventBus modBus) {
+        if (SUBSCRIBED_BUSES.add(modBus)) {
```
Cheaterpaul: do not save the bus. The caller is at fault if calling this twice.

projects/faction/src/main/java/de/teamlapen/faction/client/config/values/ColorConfigValue.java
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * If you want to use this value type in your mod, make sure to subscribe {@link ColorConfigValue#subscribe(IEventBus)}
+ * to your event bus; inside {@link Services#registerModBus(IEventBus)} if you're using {@link Services}.
+ */
+public class ColorConfigValue {
+
+    private static final Pattern HEX_PATTERN = Pattern.compile("^#?[0-9a-fA-F]{6}$");
+
+    private static final ThreadLocal<@Nullable List<ColorConfigValue>> ACTIVE_COLLECTION = new ThreadLocal<>();
```
Cheaterpaul, line 27: I do not like the thread based approach, but the simple alternatives are not really better

Cheaterpaul (changes requested):

Cheaterpaul:
For this PR I allow it, but the next PR that contains different features will be closed. If the PR adds different stuff that does not have anything to do with each other then they belong into different PRs. I have no problem when I need to handle multiple PRs.

You could have split this PR into:
1. Bugfixes
   - no need for discussions
   - i can merge it directly (or amend a few things)
2. Functional changes (the keybinding, the color config option)
   - We might discuss some things but since its technical, that should be clear
3. Cosmetics
   - For some things i have a strong feeling
   - We might need to discuss more things
   - To make this faster and not waste your time, you can pm me about new model and textures __before__ investing the time into implementing them, in case i am not convinced. (Works for other features as well)

TheGridExpert:
I'm sorry for the mess... I initially wanted to make "one fix", then it turned into quite some more. Not the first time though, this looks silly of me at this point. Will not do this again.

I usually don't DM you about such things because I try not to disturb you too much, but this case shows that I should, so, yes, will tend to warn you about my plans that are risky or related to textures/models.

(For now, will do the requested changes first, then think what to do with the blocks.)

TheGridExpert:
> Note: I changed the vampirism_factions_source_dependency parameter to true for the project to compile in the IDE, but GitHub's checks do fail anyway as a result, I assume you saw that, but will mention for those reading the thread.

ColorConfigValue now uses ConcurrentMap and doesn't saves the buses. I've looked over the alternatives for the ThreadLocal solution and, while I agree that this is not the way it's supposed to be used, other variants, as you've said, are not better, so I just made the existing one safer with a try-finally check.

The Altar of Infusion now uses a transient attribute for the slowdown and leaving aborts the ritual instead of pausing it. The initial reason for pausing it was the fact that the items are lost, so I made the altar consume them only at the moment of leveling up the player. It checks if the items are present when the ritual begins and when it is about to level-up the player. If those somehow disappear after the start (which shouldn't be possible but for commands and other mods on its own), the ritual goes on, but the user is not level-upped.

Readded the level-up sound and replaced the short choir sound with it (though it is still specifically added to each leveling station using a common method so that it isn't playing when using commands).

[8375a67](https://github.com/TeamLapen/Vampirism/pull/1610/commits/8375a674aeee58dca145f812c1f1642b1436fdf7) and [ba9d0ba](https://github.com/TeamLapen/Vampirism/pull/1610/commits/ba9d0bab6f94d8ae0446641e2065d31b78d9fffa) are commits that change the altar of cleansing and tombstones models to the old ones. Must mention that I've reexported their model files so that they're more organised than before. I've also sent the screenshots of the two variants of the blocks to your direct messages for the voting if you do one.

PR #1615, opened 2026-06-04
Language File Fixes

TheGridExpert:
## Notice

I've added the Russian translation of both Vampirism and FactionAPI (though not the guidebook for now) into the repository, but, as we've discussed previously, you can directly upload the translations to Crowdin. Withal, would like to ask if it's possible to set me there as a translation approver for Russian and Ukrainian? I could approve the translations I'm sure of (e.g. item and block names) so that those are kept. My username is TheGridExpert there, same as here, same profile picture.

## Changes

Added singular forms of faction names for better translatability. In commands, for example, faction names (e.g. Hunter, Vampire) were used as singular forms, but in other languages this often doesn't work.

Fixed parameters in block item shift descriptions; those used not to show up before.

Changed %s placeholders into their numerated variants (e.g. %1$s, %2$s) in lines where there is more than one parameter (except for coordinates and radii where there are two identical values or where those have to stay in the same order).

Corrected item shift descriptions where they were incorrect.

Renamed wall candle holders to use the prefix "wall".

Made faction totem tops use the description key of the base variant to avoid repetition that might end up being translated differently.

Renamed Cursed Sapling to Cursed Spruce Sapling.

The ID of the Alchemical Fire item was changed from "item_alchemical_fire" to "alchemical_fire" for consistency.

IMob entities now use the IDs of their normal variants so that the duplicates can be removed.

The Invisibility action had a description, but it wasn't used in the game. Fixed.

Fixed capitalization and punctuation where necessary.

Renamed the Blood Collection skill to Blood Collection and the Equipment Collection skill to Supply Collection, as "Collect" by itself sounds awkward and the word "Equipment" cannot really be used for provisions and metals.

The "Required Skills" tooltip now looks aligned with the "Faction Specifics" one:
Screenshot 2026-06-04 at 12 42 47

Renamed the translators line key from "text.vampirism.translators" to "credits.vampirism.translators". It's better not to use the prefix "text" as it's too ambiguous and we've previously sorted all "text" entries into separate categories already.


TheGridExpert:
One thing I forgot. It's better not to use placeholders inside lines where the inserted part is non-ambiguous. By that I mean situations like:

> "This consumes an Oblivion Potion" ("This consumes a '%s'")

Here the requirement is always an Oblivion Potion. It might look more consistent to do this because a translator without context might write the item name incorreclty, however, this often breaks capitalization and adds more unnecessary confusions. For example, Russian titles only have the first word capitalized, which results in the item name looking off there, as if it was inserted roughly. Better to inline names in these situations if the inserted part does not change.

Cheaterpaul:
I added you to the crowdin approver list

PR #1616, opened 2026-07-01
Hunter Crossbow Changes

TheGridExpert:
This pull request is mainly about hunter crossbows. I've just changed some things and added some new features. All features (including textures) are up for discussion, even though I've already mentioned the most crucial ones for the most part in the chat previously. All the changes are crossbow-related, though there's a plethora of them... hope it is fine.

## Double It
The Double It skill used to work a bit differently from what one would expect. I have an impression it could be a bug that appeared during a migration, though I'm unsure. It used to charge only the crossbow in the main hand while the off-hand one only appeared to be charging visually, but ended up with no quarrel. Now, provided you have the skill, the player doesn't start shooting until both crossbows are charged. The main-hand one begins charging first if both crossbows are unloaded.

The Arrow Frugality enchantment is now spread equally between crossbows when two are fired simultaneously, so you no longer end up with a different number of arrows when using semi-auto crossbows.

## Quarrel Pouch
I've removed the arrow clip and left only the pouch, but renamed it to a quarrel pouch. The pouch was missing some functionality of a bundle, so I added it and adjusted its tooltip, which now correctly displays fullness (it used to treat 64 items as the max value as the actual max storage size and the visual number were separate). Empty pouches now show a short description instead of a blank tooltip. I also used a new texture made by Zozozrob (though I tweaked it a bit to match the new quarrel textures shown below).

Screenshot 2026-06-25 at 22 35 51

There are also a couple of handling behaviors worth mentioning: you can hold right-click to spill quarrels out one at a time, and a pouch scatters its contents if the item entity is destroyed instead of losing them.

The radial screen for ammo selection now only shows the quarrels the player currently has in their inventory — better than showing all available in general, especially if more are added in the future.

Screenshot 2026-06-25 at 22 39 02

I also resolved the issue related to pose stacks in that screen... but you fixed it while I was preparing the pull request.

## Quarrel Clip
Yes, I said I removed it, but it resurrected. Basically, it's similar to its implementation in 1.12.2, though not exactly. It is crafted from 16 normal quarrels or via the recipe below if someone wants to save time (though resource-wise it's the same unless you craft quarrels from vanilla arrows). You simply load it as ammo and 16 quarrels are inserted into the crossbow. Initially, I wanted to make it contain just 12, but Mike suggested 16, so it's 16.

Screenshot 2026-06-25 at 22 47 27

## Quarrels
In addition, I've drawn new textures for quarrels; they're now different from vanilla arrows and more like real-life bolts. Only the tipped arrows have a tint applied now. They also have special textures in the form of an entity.

Screenshot 2026-06-25 at 22 44 34

The entity model is two pixels shorter than that of the vanilla arrow by the way, that's not just a reskin.

Screenshot 2026-06-26 at 00 11 58

I have also renamed quarrels in the code. They used to be called "crossbow_arrows" in their IDs and some other places, which could be confusing in the long term. One config "arrowVampireKillerMaxHealth" was renamed to "quarrelVampireKillerMaxHealth" as well. I hope this doesn't make it more difficult to review the code changes...

## Spitfire Quarrels
Spitfire quarrels now look as if they're burning in alchemical fire rather than the regular one — just a visual detail.

Screenshot 2026-06-26 at 00 20 14

What is more vital is the shape of the alchemical fire pool they create on impact. A single block felt like too few, so I've made it more random, as if the charge splashes all around. It covers more area and feels more useful than before. Each small puddle on the screenshot is one quarrel shot.
> I should note that the code I wrote for it is universal; it simply creates a splash-like pool. It can be used for similar things in the future that spit alchemical fire. I thought of adding the alchemical cocktail I have in Vampire's Delight to the main mod, but it would need to be implemented in a more balanced way, so I postponed it for now.

Screenshot 2026-06-26 at 00 23 18

## Teleport Quarrels
Teleport quarrels previously didn't teleport the player when hitting an entity. This might be overpowered, not sure, but I made it teleport the shooter behind the entity that was hit. It's meant to be combined with backstabbing weapons and skills that increase damage when you hit from behind. It should be fun to use; just hope it won't be annoying.

To make this counterable, I've also added teleportation particles that appear while the player is charging the crossbow to warn the enemy player that they are preparing to teleport.

In addition, teleport quarrels now load the area around themselves within a small radius, like ender pearls do on the latest MC versions, which makes them work better.

## Heavy Quarrels and Heavy Quarrel Clip
Heavy quarrels are a new variety of quarrels with greater piercing capabilities, larger damage, and more knockback, but they're more affected by gravity and slower.

Screenshot 2026-06-28 at 16 53 50

Heavy quarrel stats:
- +0.5 base damage
- 60% result damage multiplier (actually less because of lesser speed, actually around 40%)
- 40% more knockback
- 20% slower
- 50% heavier
- 50% more accurate (less afected by wind)
- 2 extra pierce levels (same as with piercing II enchantment, it stacks)
- 33% longer charging

Since heavy quarrels aren't exactly the same as effect quarrels, I've decided to make a special quarrel clip for them. That said, it doesn't contain 16 quarrels, it holds 8 instead. That should be a bit better balance-wise. It can be crafted from 8 heavy quarrels or via the recipe below.

Screenshot 2026-06-28 at 16 53 53

Heavy quarrels also have a special entity model that is darker and longer than the default one.

Screenshot 2026-07-01 at 13 10 59

In addition, I've also included a tooltip for quarrel clips. It simply shows what kind of ammunition the cartridge carries.

Screenshot 2026-07-01 at 12 41 08

Semi-auto crossbows now also work with the ammo selection screen — you can choose which quarrel clip you want to use out of the two.

Screenshot 2026-07-01 at 12 44 07

## Precision

Precision is a new enchantment with three levels for hunter crossbows. Each level reduces inaccuracy and gravity by 30%, which makes it easier to use your crossbow essentially as a sniper rifle. I should note that quarrel gravity is reduced only when the quarrel has enough velocity. I couldn't scale its speed instead, since that would also make the quarrel deal more damage due to the way the game calculates it, so this should be fine.

This enchantment is only obtainable after completing a task for a hunter representative called "Sights Calibration". There are three stages; each depends on your hunter level and the Precision level you'll receive. For level 1 you need to kill 15 vampires with a crossbow; for level 2 the number rises to 30; and for level 3 it's 45.

Screenshot 2026-07-01 at 13 02 59

There's a problem with the task screen, though: it doesn't display which enchantment you're going to receive, so it may be confusing. I'll probably come back to that later as this PR is about crossbows.

## Spacing
On top of that, I've added spacing between the fired quarrels. Not much, but enough to keep them from colliding. Double crossbows have a small spacing, while crossbows held in two hands have a larger one. Here's how shooting with two double crossbows looks. I applied Precision III to make the difference more visible, so there's no inaccuracy-driven scatter.

Screenshot 2026-07-01 at 15 14 03

## Hurt Timer
Previously, double crossbows were essentially useless, since shooting two quarrels into one enemy was treated as spam damage by the game. Now up to 3 quarrels can deal damage without being blocked. So shooting from a double crossbow, or with two single ones, now makes sense — though using two double ones isn't as efficient, as only 3 quarrels will actually land. (I've thought of making a player able to wield either double crossbows or two crossbows simulteniously because of skill limits, wrote about that in the chat, so this might be redundant if that change is applied).

## New 2D Textures

Basic and enhanced crossbows in the 2D pack received new textures. Initially, these were made by me, but Zozozrob tweaked the shading so that they appear less plain when scaled. I haven't finished textures for semi-auto crossbows yet; I'll make them a bit later, once I come up with a good variant.

Screenshot 2026-07-01 at 19 12 58

> In-game appearance

Screenshot 2026-07-01 at 19 14 42

> All pulling stages

Note: enhanced crossbows don't just have more pixels, they're actually bigger in the game; the pixel ratio is all correct, don't worry, I considered that.

They now all display the pulling process. At least in the 2D variants. 3D models behave the same as before; I didn't manage to edit their models to look good, as if they're actually being pulled.

I also currently have an idea to remake all the current 2D textures so they fit the game better and use them as the default, while moving the existing 3D models into a separate resource pack for those who prefer them.

## Advancements

I wanted to add new advancements related to hunter crossbows, so here they are.

Screenshot 2026-07-01 at 20 41 38

> A fairly popular phrase related to shooting, so I decided to use it.

Screenshot 2026-07-01 at 20 41 41

> Many weapon designers use the name "Mark" followed by a number, it's just a common practice among engineers. However, many associate it with Iron Man specifically, so it also works.

Screenshot 2026-07-01 at 20 41 46

> This one is pretty literal — semi-automatic crossbows aren't *that* far from guns. Even so, it's actually a reference to The Amazing Digital Circus, that phrase was from the episode where the main characters shot each other.

Screenshot 2026-07-01 at 20 41 50

> This one already exists in vanilla, but it required shooting through 5 mobs. This one requires 7, which is achievable by using heavy quarrels and Piercing V on top.

Screenshot 2026-07-01 at 20 41 56

> The quote comes from Team Fortress 2, where the Sniper says it. The player needs to shoot a mob using a crossbow with Precision III. I picked 128 blocks specifically because it's a power of 2 and the mob still renders on screen even at this distance; at 150 it's so far that the game stops rendering it (at least for me) (this number also looks cool).

## Item Descriptions

Updated the teleport quarrel's on-impact description to clarify that it teleports the shooter behind the mob that was hit.

The normal and heavy quarrels now have full shift descriptions instead, to explain what they do more clearly.

I've also added shift descriptions for crucifixes and revised the skill descriptions related to them to convey what they do more accurately.


TheGridExpert:
Merged the changes on dev into my branch. There *might* be some problems, but shouldn't be.

TheGridExpert:
Not sure why I didn't do that initially, but I added separate 16x16 textures for enhanced crossbows that are used inside the inventory. Spears work the same way in vanilla. I'll probably do that to all weapon textures bigger than 16x16, at least those that I make.

Screenshot 2026-07-03 at 22 45 13

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/world/items/component/QuarrelPouchContents.java
```diff
@@ -2,52 +2,48 @@
...
-import org.jspecify.annotations.Nullable;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
-import java.util.stream.Collectors;
 
-public record QuarrelPouchContents(List<ItemStackTemplate> items) implements TooltipComponent {
+public record QuarrelPouchContents(List<ItemStack> items) implements TooltipComponent {
```
Cheaterpaul: Should work with ItemStackTemplates instead of ItemStacks. Maybe a merge issue

projects/vampirism/src/main/java/de/teamlapen/vampirism/VampirismMod.java
```diff
@@ -71,7 +71,7 @@ private void checkEnv() {
     }
 
     private void processIMC(final InterModProcessEvent event) {
```
Cheaterpaul: Can you move that into the CommonServices registration

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/world/items/crossbow/QuarrelHandler.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * @return all registered clip items (carrying a {@link ModDataComponents#CONTAINED_PROJECTILES} component)
+     */
+    public static Set<Item> getClips() {
+        if (clips == null) {
+            clips = BuiltInRegistries.ITEM.stream().filter(item -> item.components().has(ModDataComponents.CONTAINED_PROJECTILES.get())).collect(Collectors.toUnmodifiableSet());
+        }
+        return clips;
+    }
```
Cheaterpaul, line 45: Why not do same as above and retrieve it in the mod events

Cheaterpaul (changes requested):
Looks good.

What is the reasoning of adding the heavy quarrel ones?

Cheaterpaul:
I tested the precision enchantments, and they seem to strong.

I would suggest reducing the precision from 30% per level to 10% per level and remove the enchantments from tech crossbows

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/world/items/component/QuarrelPouchContents.java
TheGridExpert: It was a merge issue, resolved

projects/vampirism/src/main/java/de/teamlapen/vampirism/VampirismMod.java
TheGridExpert: Moved

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/world/items/crossbow/QuarrelHandler.java
TheGridExpert: Had a problem because data components were not registered at the moment I tried to access them, but I found an event that runs after they are registered, so it should be fine. Resolved

TheGridExpert:
The reason I added heavy quarrel clips is simple: variety. Allowing semi-auto crossbows to use effect quarrels is OP, I agree with that, but I can't deny that it makes them plain compared to normal crossbows. Technically, heavy quarrels are not effect quarrels; they're just a different kind of quarrel, so I wanted to make an exception here to give players some interesting stuff to use. The heavy quarrel load time is 33% higher, and on top of the semi-auto reload time and the fact that the clip only contains 8 arrows, it feels far slower, so it should be a fair trade-off.

As for the precision enchantment, though, I decided to change the values in a different way rather than just lowering them. The initial concept is to make you shoot more accurately so that your quarrels hit the same spot precisely, while the gravity effect is a separate side effect that was very questionable. I kept 30% for the accuracy factor (it's not really something too unbalanced on its own, I think), but made a separate one for gravity. 10% felt like too little, so I set it to 13.3% (40% on level 3) so that it's a bit more noticeable, but not as critical as with 30% per level. I hope this solution is fine. The values are taken out separately, so you can change it to 10% if 13% is too much. And yes, semi-auto crossbows cannot use it anymore; I removed them from the list.

Cheaterpaul (approved):
> found an event that runs after they are registered,

nice find



We might need to adjust the balancing of the precision and the heavy quarrels, but that is something for balancing testing with mike.

PR #1619, opened 2026-07-17
Biome Adjustments

TheGridExpert:
## Vampire Forest Vegetation

I wanted to make the vampire forest feel thicker and creepier, that's why I made more grass, cursed roots and vampire orchids generate there. It's not just about adjusting the values; now there's also a decorator for vampire trees that creates tall grass around them so that it looks more organic and not just scattered all over the place.

Screenshot 2026-07-16 at 15 09 48

Additionally, I've adjusted the Velmorra biome a bit. I assume the difference in tints appeared as a result of us editing two biomes in parallel, so I made Velmorra's foliage, fog and water tints match the vampire forest's. The differences are that Velmorra has the same generation as before (didn't touch it, just organised a bit), there are no ambient sounds (I suppose that was intentional), and the leaf color is darker than normal (kept it darker to make the place look more vibrant). The grass color is the same as in the vampire forest because it has to match the cursed grass color. It looks good when grass matches its herbage.

## Dark Stone

I wasn't exactly fond of dark stone generating like ore below vampire forests, so I came up with a different approach.

There is now a layer of dark stone below cursed earth (5 blocks on its own). Below that layer is noise that creates some sort of small roots to show that the forest is corrupting not just the dirt but everything below as well. The following screenshot was taken below the ground with vanilla stone removed.

Screenshot 2026-07-16 at 15 13 15

Those roots are quite short, and so they don't go too deep. However, one might want to dig up more dark stone, since many players use it for builds. That's why I also added giant roots as a feature. Those can go down to y 0 and can vary in width.

Here's how it looks if everything but the dark stone is cleared.

Screenshot 2026-07-16 at 15 12 11

I must note that the roots can pierce caves, water and lava lakes, but they cannot go through structures such as the crypt and ore veins.

### Usage

Now cobbled dark stone can be used for crafting stone tools and several blocks like a furnace (similarly to basalt and deepslate). I assume that would make sense if someone wants to survive only in the vampire forest, considering that the top layer of stone is replaced with dark stone.

## Crypt

Added a separate structure class for the crypt. Now the top piece is generated as `beard_thin` while the rest are just buried. This fixes a bug where the crypt deformed the land near itself with its underground part.

The staircase is also no longer broken. I did this by directly editing the raw nbt files of the staircases. For some reason this setting is not present in the jigsaw block itself, but you can add or edit a "joint" field and set it to "aligned" inside the jigsaw block nbt. This causes it to not rotate while being placed.

There was also another bug where vampire forest vegetation (grass, flowers and roots) replaced the floor of the crypt church. There was a modifier that almost all vanilla plants have that checks if the target block it air. I added it for the mod's plants, now it works correctly.

Screenshot 2026-06-24 at 19 02 28

## Technical Note

I adjusted the Integrations class as it previosuly made vampire forests not contain cursed grass and earth. I hope this is fine.


Cheaterpaul:
The giant root look even good in velmorra :D

PR #1625, opened 2026-07-30
Vampire-Related Mechanics

TheGridExpert:
## Passive Leveling

Vampire players can now level up passively up to level 4; it takes 5 days per level-up.

This needs more context on how and why it's implemented, but briefly — it's not meant to make leveling up easier. There are other reasons I thought this could be added, and they might not be so obvious.

The main reason is to push players towards leveling up in the first place. I've noticed that some new players stay at level 1 and don't do anything, but then complain that they're vampires and don't even burn in the sun. There is a bar in the faction menu specifically for level 1–3 vampires that explains that you need to use the altar of inspiration, and how passive leveling works. Not the best visual design... but I hope it's fine. Essentially, it gives the player directions and tells them what to do.

Screenshot 2026-07-26 at 18 37 50

Another reason, besides giving directions, is to hint at what levels 1–3 mean. There are already references to this in the level 4 skill ("No Longer A Fledgling"), but this makes it clearer. You're essentially not a full vampire yet — that's why you don't burn in the sun, and that's why you grow passively.

The duration of 5 days is configurable, but I'd like to hear the balancing team's opinion. It's supposed to be a period of time that is too long to wait out just to save resources: if you want to level up, you'd rather craft an altar and gather blood, while those who don't want to level up and just want the basic buffs without as many debuffs will eventually start taking sun damage. I'd also mention that sleeping doesn't count towards that timer.

## Sun Sensitivity

When a vampire is under the sun, they burn — so why wouldn't their eyes? This is a purely visual thing, but I thought it would be more immersive to have a sun blinding overlay when looking directly at the sun as a vampire.

There are definitely people who wouldn't want their monitor flashbanged because of their curiosity, so I've added an option that makes the sun blinding overlay black. It's located in the vanilla Accessibility menu, since that appears at the start of the game and is a more visible spot for people who are new to mods in general. I can move it into the configs if you'd like (though there's a separate line there to remove the overlay entirely).

Screenshot 2026-07-29 at 17 20 01

Here's a recording how it looks in game.

https://github.com/user-attachments/assets/1e1b0f19-e422-44fb-b3ba-9f5ccdc21963

It's better to look at it in game, as the video doesn't fully capture it, perhaps. The overlay also has a lens flare effect; I tried to imitate how it looks in real life to make it more vibrant.

## Sanguinare

The Sanguinare Vampiris effect already had stages, but they only applied hunger, so I added a few more effects that appear closer to the end. When 25% of the duration is left, the infected player gains a new sun sensitivity effect. The idea behind this is that your eyes become sensitive to the sun before the transformation finishes, so it should signal that not much time is left.

Screenshot 2026-07-27 at 00 18 08

(I hope the texture is fine — this effect only appears as part of Sanguinare Vampiris, and I didn't manage to make a better one at least for now :p.)

When 10% of the time is left, you gain the Exposed effect. Under normal circumstances, this lasts around 1.5 minutes (assuming the infection takes approximately 15 minutes). If you don't rush into fighting (which you shouldn't do anyway, since Sanguinare disappears on death), you should be fine.

Sanguinare is now also split into 4 stages. The hunger brings stage II, the sun sensitivity stage III, and the exposed stage IV. The stage number is displayed in the inventory as the effect's amplifier. It does nothing on its own — it just indicates that the infection is progressing.

A small thing, but all the effects added by Sanguinare have their particles hidden so as not to clutter the screen.

The particles themselves have also been changed. The previous bat ones were too big and took up too much of the screen for the duration of the effect, so I added more sprites: 3 bubbles taken from the same texture as the bat. Now it should look more interesting.

https://github.com/user-attachments/assets/c369a6fa-a952-4216-b890-5e421c7a0321

I also fixed a bug where Sanguinare Vampiris disappeared after the player rejoined the world. It was caused by the lack of serialization for the infection status data.

## Sun Burning

Previously, vampires were just knocked back and forth by the sun without actually burning (not counting the ash particles), which looked quite humorous to me, so I decided to change that. I didn't touch the damage or the way it's dealt — I only added a visual burning effect. It's identical to the vanilla fire, except that it's more yellow and isn't visible in first-person view. I also tried tinting mobs yellow on top of that, but it looked weird, so I didn't implement it.

https://github.com/user-attachments/assets/f36c3bd9-733d-4bf1-bff3-823844bac291

Note: converted mobs stored the sun burning data only on the server, so I had to adjust their code.


TheGridExpert:
## Passive Leveling

Passive leveling no longer depends on time. Instead, as Paul and Rand suggested, it was changed to work off the blood consumed by the player. The requirement is 2.5 times bigger than for the Altar of Inspiration — I think that should be fair, though it can be tuned up or down later. The mechanic accounts only for the blood the player drank from a mob and consumed, not poured into a bottle.

The max level achievable passively is now 3.

## Restrictions

DBND is now restricted to vampires of level 4 and higher. Configurable.

All vampire swords and hunter axes now share identical level restrictions: level 4/9/14 for normal/enhanced/ultimate tears respectively.

Hunter armor now requires the same skills to be worn as it does to be crafted.

## Sun Blindness

The sun blind overlay intensity now depends on the vampire level. For a human with Sanguinare and a level 1 vampire it's 25% max, for level 2 it's 50%, 75% for level 3, and 100% for level 4 and higher.

Now, instead of inversion, there's an option to tone the blindness down with a slider in the accessibility menu.

I've extended the OptifineHandler class — it's now called ShaderHandler, as it looks for the Iris mod as well. The blood vision is turned off only if OptiFine shaders are active, the way it was before. It didn't break for me with Iris, so I assume it's fine? The sun blindness overlay is turned off if either Iris or OptiFine has active shaders, but it can be brought back with a separate config if needed: it actually looks good with some shaders, surprisingly. Complementary shaders, for example — so it's a matter of taste and shaders.

## Notes

I did not add flame particles to the vampire burning effect, since it looked overloaded and simply... bad. The existing texture is more than enough for that, at least in my opinion.

I didn't add that Skyrim overlay in any way, it would be terrible in the game.

projects/vampirism/src/main/java/de/teamlapen/vampirism/common/core/ModItems.java
```diff
@@ -108,34 +109,34 @@ public class ModItems {
...
+    public static final DeferredItem<HunterCoatItem> HUNTER_COAT_FEET_ENHANCED = ITEMS.registerItem("hunter_coat_feet_enhanced",  props -> new HunterCoatItem(ModArmorMaterials.ENHANCED_HUNTER_COAT, ArmorType.BOOTS, IItemWithTier.Tier.ENHANCED, List.of(HunterSkills.WEAPON_TABLE, HunterSkills.MASTER_CRAFTSMANSHIP), props));
+    public static final DeferredItem<HunterCoatItem> HUNTER_COAT_FEET_ULTIMATE = ITEMS.registerItem("hunter_coat_feet_ultimate",  props -> new HunterCoatItem(ModArmorMaterials.ULTIMATE_HUNTER_COAT, ArmorType.BOOTS, IItemWithTier.Tier.ULTIMATE, List.of(HunterSkills.WEAPON_TABLE, HunterSkills.MASTER_CRAFTSMANSHIP, HunterSkills.ARTISAN_CRAFTSMANSHIP), props));
+
+    public static final DeferredItem<HunterArmorItem> HUNTER_HAT_TALL = ITEMS.registerItem("hunter_hat_tall",  props -> new HunterArmorItem(ModArmorMaterials.HUNTER_HAT_TALL, ArmorType.HELMET, List.of(), props));
+    public static final DeferredItem<HunterArmorItem> HUNTER_HAT_BROAD = ITEMS.registerItem("hunter_hat_broad",  props -> new HunterArmorItem(ModArmorMaterials.HUNTER_HAT_BROAD, ArmorType.HELMET, List.of(), props));
```
Cheaterpaul: I would only list the skills as required which are required to craft them, which is only the most advanced skill here. To unlock that skill you need to unlock the first skills as well, a quick lookup in the skill tree will show you what you need.

Otherwise the tooltip is cluttered with unnecessary data and the skill check does unnecessary checks

Cheaterpaul (changes requested):

Cheaterpaul (approved):

PR #1630, opened 2026-08-05
Hunter Mechanics

TheGridExpert:
## Power
 
Hunter crossbows now support Power (I–V).
 
As a consequence, I had to add quarrel entities to the `minecraft:arrow` tag for that to work. The only thing this seems to affect is that quarrels can now destroy decorative pots and stalactites... and, well, take part in advancements.
 
## Dual Wielding Skills
 
Now there's a separate skill called `Dual Wielding` for using two crossbows at once, while `Double It` is used to craft double crossbows. This is meant to avoid being able to shoot 4 quarrels at once.
 
Now crossbows require the following skills:
 
| **Crossbows** | Basic Single     | Basic Single                | Enhanced Single      | Enhanced Double                 | Basic Semi-Auto      | Enhanced Semi-Auto    |
|---------------|------------------|-----------------------------|----------------------|---------------------------------|----------------------|-----------------------|
| **Skills**    | Basic Technology | Basic Technology, Double It | Master Craftsmanship | Master Craftsmanship, Double It | Master Craftsmanship | Artisan Craftsmanship |
 
Raised requirements for semi-automatic crossbows as well.
 
Also updated the icons of crossbow-related skills.
 
Screenshot 2026-08-04 at 23 07 49

## Hunter Axes
 
Removed knockback from hunter axes, as it was suggested by quite a few people.
 
## Detrition Deferment
 
A new skill in the hunter weapon branch that is more of a "comfort feature" than something that will make you deal more damage. This skill makes your armor not break completely. Instead, there will be a 60-second period of the armor keeping its protective capabilities. After the time runs out, it still doesn't break, but stops protecting the wearer until repaired. There's a timer on its icon to indicate how much time you have left.
 
Screenshot 2026-08-05 at 11 54 20

I'm not sure where to put this, to be honest, as the general idea seems to be better for beginners. I may move it higher up. Currently, it's after the `Master Craftsmanship` skill.
 
Screenshot 2026-08-04 at 23 29 54

This is how armor looks when under the effect of this skill. It doesn't show the exact time, but demonstrates the approximate percentage of it. Made it blue, as other colors were either not as visible or looked weird.
 
Screenshot 2026-08-04 at 23 30 42

As the effect runs out, this pie turns into a cross.
 
There are two new stopwatch sounds that are played when the skill effect activates and when it ends.
 
## Item Requirements
 
All hunter armor and weapon crafting recipes now share skill and level requirements as if you used those items. Lava required for crafting now has a specific scaling: 1/3/5 lava for Normal/Enhanced/Ultimate tiers respectively, not random.
 
Crafting hunter axes no longer requires skills; it requires the same levels using them does.
 
Alchemical cauldron recipes now also have fluid unlock criteria. Also cleaned up some of the weapon recipes.
 
## Heavy Quarrel Rework
 
Heavy quarrels are now more of a situational kind of quarrel. They are now 2 times heavier than normal quarrels, and their base damage multiplier is a bit smaller. However, they now share a new ballista mechanic. The way it works is pretty simple, but it might be confusing for some, so I made a few schemes.
 
When calculating the ballista damage multiplier, the quarrel takes the lesser of the blocks it flew upward (increased by 20%) and the blocks it descended, then divides that value by the quarrel's maximum height it can fly vertically up. Then, this value (between 0 and 1) is multiplied by the maximum ballista multiplier (currently 3 — consider that you wouldn't get to it under normal circumstances. Shooting yourself is the most efficient though... yeah...).
 
sceme1

scheme2

scheme3

Essentially, if you're above the target vertically, you'll get a more powerful shot than if it were higher than you.
 
For some real value, one would need to shoot at around 60 degrees, which is quite challenging.


TheGridExpert:
(I adjusted my name in credits a bit, just used the full username in the mod description. I hope it's fine?)

supermike1999:
Alright:

**Power:** Looks good to me

**Dual Wielding Skills:** Testing out dual wielding double crossbows, I found they aren't as op when you do land hits with them, considering I'm hitting a sitting player, the likely chance of landing hits on a player running about is even less. 

so I say these skill can be separated if you so choose.

The "Raised requirements for semi-automatic" should be reduced down to 
Basic Semi-Auto --> Basic Technology
Enhanced Semi-Auto -->Master Craftsmanship

*I find needing to go all the way to Artisan Craftsmanshipto unlock the Enhanced Semi-Auto is a bit much for one crossbow type and they should be stuck with the other crossbow types when unlocking them (Basic crossbows requiring Basic Technology  while the enhanced version require Master Craftsmanship

**Hunter Axes**
Fine to me personally, as I usually just remove the knockback with a grindstone. Only thing I can think of is I know some players did like having the absurd knock back, so they might be a bit disappointed in it being gone.

**Detrition Deferment**
No
It doesn't make any sense to have something like this when unbreaking, mending & extra armor sets exists plus i find it very hard to break armor in a pve world unless you happen to go afk and let a mob smack you till the armor breaks. 

This is also, without a doubt, a very broken mechanic in any pvp scenario,  given how after testing you can easily throw down one xp orb to reset the timer. Even if a cooldown was added for it before it activates again, all you would need is to swap it out with another set to allow the timer to come off cooldown. 

This is completely unbalanced and should be removed


**Item Requirements**
Looks good to me

**Heavy Quarrel Rework**
I don't know what more to say here, I still personally don't see a reason for these quarrels to exist.

Testing them out with the changes, their slight damage increase is still not noticeable, even at a elevated advantage vs another player or mob, the damage increase I find isn't much. 

Given the return of power being applied to the crossbows and the already existing piercing enchant, I  still don't feel to feel worth using over normal quarrels and think they simply should be removed.

TheGridExpert:
## Detrition Deferment

I can nerf this skill so that it works in a different way. When nearly breaking, the armor doesn't break completely, but instantly loses its protection. When you repair it, there would be a 60-second timer until it starts protecting again.

## Heavy Quarrels

Again, as I've said, they're for fun. If you tried shooting them with low angles — yes, the damage increase is unnoticable, that's not what they are meant for. I will not remove them, but if they're not widely used everywhere — this is fine, they're not designed for that.

TheGridExpert:
Though, to be honest, this kind of nerf for Detrition Deferment makes it very useless. I would rather make it not recover after 60 second, but after you repair it *fully*. This way, it should make it more usefull without breaking PvP much.

Cheaterpaul:
### Detrition Deferment

- I have the feeling that the Detrition Deferment skill is only useful for beginners that are not yet able to get the mending enchantment. And then you would never risk it to loose your armor.
- It confuses the opposing players that see the armor, but it may have no effect.
- Other inventory screens do not display the items (and other problems i noticed with werewolves)

If you want the effect of a broken armor that can be repaired, I thing an all around solution that has a similar effect and reduces incompatibility and confusion might be the following: I would suggest that the armor is converted into a broken armor item (drops into inventory or in the world if no space is left (too bad for the owner)), the broken armor item then can be repaired into the original item (keeping all data components), with low durability. Then the player can decide what to do with it (repair, mending, etc...).

### Rest

Is fine

TheGridExpert:
## ~~Detrition Deferment~~ Near-Breach Reforging

Screenshot 2026-08-09 at 12 12 17

Remade this skill. Now it basically doesn't let your armor to break fully, as Paul suggested. One can repair it back (only) on an anvil.

I initially wanted to move it higher, but left where it was beause it made the branch look odd. I'll come back to that later.

Screenshot 2026-08-09 at 12 12 01

The shattered armor looks like the original, but with a break overlay on top. Was quite... a quest to implement (left comments there for clarity), but it works even with 3d resourcepacks now.

projects/faction-api/src/main/java/de/teamlapen/faction/api/factions/skills/SkillProperties.java
```diff
@@ -108,6 +113,11 @@ public SkillProperties<T> withDescription() {
         return this;
     }
 
+    public SkillProperties<T> withDescription(Supplier<?>... args) {
```
Cheaterpaul: When i see it correctly this methods is no longer used. So the SkillProperties changes can be reverted.

projects/faction/src/main/java/de/teamlapen/faction/common/factions/skills/Skill.java
```diff
@@ -33,11 +32,14 @@ public class Skill<T extends IFactionPlayer<T> & ISkillPlayer<T>> implements ISk
...
 
     private final Map<Holder<Attribute>, SkillAttributeHolder> attributeModifierMap;
+    private final Supplier<@Nullable Component> description;
 
     public Skill(SkillProperties<T> properties) {
         this.builtInRegistryHolder = ModRegistries.SKILLS.createIntrusiveHolder(this);
         this.descriptionId = properties.effectiveNameId();
         this.attributeModifierMap = properties.attributes();
+        List<Supplier<?>> descriptionArgs = properties.descriptionArgs();
+        this.description = descriptionArgs.isEmpty() ? () -> this.components().get(FactionDataComponents.SKILL_DESCRIPTION) : () -> Component.translatable(this.descriptionId + ".desc", descriptionArgs.stream().map(Supplier::get).toArray());
```
Cheaterpaul: Same here, it is no longer used.

One reason I moved to the property based initialization (like for the faction proeprties or the item properties) is that all values are now fixed and do not need on demand object creation (e.g getName() -> Components.translatable(getDescriptionId())). That why i would prefer such initializer to be fixed and not used with suppliers or something similar (if it makes sense and actually works). FYI: I just added a ConfigCodec and ConfigComponent to let the config values be serializable.

projects/vampirism/src/main/java/de/teamlapen/vampirism/client/core/ModItemsRender.java
```diff
@@ -46,22 +49,40 @@ public static void registerConditional(RegisterConditionalItemModelPropertyEvent
...
+            return;
+        }
+
+        buffers.clear();
+        previous.forEach((type, buffer) -> {
+            buffers.put(type, buffer);
+            if (type == Sheets.translucentItemSheet()) {
+                buffers.put(crumbling, new ByteBufferBuilder(crumbling.bufferSize()));
+            }
         });
```
Cheaterpaul: If we use mixins anyway, can we just directly inject a put call in the the SequencedMap creation.

Cheaterpaul (changes requested):

Cheaterpaul:
And i forgot. Can you put the Near-Breach Reforging skill as option to the first level of the skill branch instead of this late, as it is mostly usefull for the early game.

TheGridExpert:
Should be it now 👍

Cheaterpaul (approved):

PR #1634, opened 2026-08-22
Skill Segment System

TheGridExpert:
## Skill Segment System

I have remade the previous skill node and tree system, replacing it with a new segment one. The main reason for that was the previous variant being quite quirky and hard to edit, as it stored node references in one file. Therefore, if one wanted to restructure the tree, they would need to modify all skills in the branch.

Segments (picked a new name to make clear it's different from what it used to be) are basically chains that connect together to form a tree instead of relying on one reference file. A segment holds the skills it contains, the segments it locks, its parents and its priority. Most of those are self-explanatory; priority will be explained later, nothing complex.

It has the following structure:
```
{
  "tree": "modid:tree1",
  "skills": [
    "modid:skill1",
    "modid:skill2"
  ],
  "parents": [
    "modid:parent1",
    "modid:parent2"
  ],
  "locking_segments": [
    "modid:locked_segment1",
    "modid:locked_segment2",
    "modid:locked_segment3"
  ],
  "priority": 1
}
```

A skill tree graph (also a new term, so it doesn't get mixed up with the old tree data) is built when data is ready on both the server and on clients. It's a class that collects all segments together into a tree, called SkillTreeGraph. SkillTreeGraphs, though, is the one that handles its caching and creation. Broken segments are handled, if one is corrupted, all of its children are abandoned.

SkillTreeLayout is a new class. I basically just split the positioning of the segments into a separate class that later gets accessed on render. Also edited the way segments are positioned in the skill tree visually. Now they look a bit more packed together and branches that the main one is split into are centred. The priority value mentioned previously determines the horizontal position of the segment relative to its siblings: the lower the value, the more left; defaults to 0.

What all of that actually gives:
- One segment can have multiple parents. This resolves the weird look of the right branch of the hunter tree, at least its top.
- One skill tree can have multiple roots. I dunno what this is for, but it's possible. Maybe it will be useful somewhere, considering FactionApi is now separate.
- Addon mods can inject their own skills into a skill tree.
- Skill trees look cleaner.

Skill trees can now also specify what other trees they come after, just in case another mod wants its tree to come after the level one, for example.

## Skill Tree Changes

The vampire skill tree is unchanged and looks the same. The hunter one, however, has had some changes.

1. The left branch gets split into two: diffusers and crucifixes. I decided to do that as you could previously unlock purified garlic before being able to use diffusers... which is odd. This variation should look better.
2. I've made new icons for the skills in the middle tree. I like how the book one came out :3
3. Modified certain skill costs to balance the tree unlockability at least a bit with vampires:
    - Improved Diffuser: 2 -> 1
    - Multitask Brewing: 2 -> 1
    - Master Brewer: 3 -> 2
    - Potion Resistance: 2 -> 1
    - Armorbound Swiftness and Highvault: 2 -> 1 each
    - Near-Breach Reforging: 2 -> 1
    - How to Actually Use a Stake: 2 -> 1
    - How to Actually Use an Axe: 3 -> 2

Screenshot 2026-08-22 at 12 32 56

## Skill Unlocking and Forgetting

One now needs to hold a skill for 8 ticks to unlock it. Looks a bit more interesting than just clicking. However, now forgetting individual skills is also possible with by holding the right mouse button takes 16 ticks.

To forget a skill, you need to be either in creative mode or have a potion of oblivion with enough portions. One portion lets you forgive one skill, 24 per bottle. If one drinks it, not all skills get forgotten as before, but the same number as the number of portions left in the bottle.

There's no need to forget the last active skill: if one with active children is reset, its children will as well (if there's no other parents those can depend on).

The only thing is that I did not make any UI for the oblivion portions. I planned to change the GUI before, so I'll come back to this later.


projects/faction-api/src/main/java/de/teamlapen/faction/api/factions/skills/ISkillSegment.java
```diff
@@ -0,0 +1,50 @@
...
+     * The relation is stored one way only, so a mutual exclusion has to be declared on both segments.
+     */
+    List<ResourceKey<ISkillSegment>> lockingSegments();
+
+    /**
+     * Determines the position of the segment in relation to its siblings, lower values coming first.
+     * For instance, if one segment has priority 0 (default) and the other one has 1, the latter one will be positioned
+     * second.
+     */
+    int priority();
```
Cheaterpaul: would it make sense to use dedicated "before" and "after" `ISkillSegement` values for better targeting instead of relying on arbitrary integer?

Cheaterpaul (changes requested):
Nice improvement.

Way easier way to configure the tree and more versatile.

I have two things that I am not fond of.

### Skill point cost reduction

They are to much, now a hunter can unlock almost all skills, while having even more skills than vampires. I'd rather see that vampire player can unlock fewer skills or rather have more skills to choose from.

The player should choose a way and not just unlock whatever.
I want them too choose a way. They should be able to max out a maximum of 2 sub-trees with some extra and not more.


### Skill reset

I do not like the new way of oblivion charges. Now you can basically lock and unlock skills freely, and that very quickly. It feels you no longer need to think about what you want to unlock, because it is so easy to unlock, forget and unlock other stuff.

What was your goal here, did you wanted exactly that?

I'd rather want the player to make decisions and live with them and not just revert it free whenever. (Yes first you need a potion, but then you have 25 or 10 charges). Additionally the need to reset all skills just to re-skill some skills is not a bad decision either.


(As always if I do not definitively say no, we can discuss it.)

TheGridExpert:
## Skill Point Reduction

Okay, I will probably just revert my changes. We'll probably add some new skills for vampires eventually, so it should be fine.

## Skill Reset

Got you, that's okay. I initially planned the individual revertion soley for creative, but then just got an idea to add it for survival as well. The reason I did this was that survival players normally get a ton of potions of oblivion while looting hunter outposts and some other structures, so they are not rare at all, even too common. I'd rather revert this decision to making players reset all kills with a whole bottle, but would also decrease the potion spawn chance (because you clearly don't need more than a few in one world). For creative I might leave the individual revertation so that we or other people could use it for testing, if you don't mind. It's still faster than forgetting all the skill tree.

Cheaterpaul:
> For creative I might leave the individual revertation so that we or other people could use it for testing, if you don't mind

Yes, for creative that is not a problem

> but would also decrease the potion spawn chance

Sounds good

TheGridExpert:
Okay, made the following changes:

- Skills can now be forgotten individually only in creative mode.
- Oblivon potions spawn not as often in hunter outposts as crypts.
- Rollbacked skill costs to what they were before.
- The skill segment priority value is replaced with a segment placement value. It looks like this in the data:
```"placement": {
    "type": "after/before",
    "segment": "modid:relative_segment"
  }
```
The type field is an enum that can be either "after" or "before". I initially made it extensible, but then scrapped that as... not sure what other values could possibly be added.

PR #1637, opened 2026-09-02
Potion and Oil Adjustments

TheGridExpert:
## Potion Recipe Cost

I have always found the amount of resources required to brew hunter potions unnecessarily inflated. Fitting a stack of glowstone blocks into a few bottles is quite... odd. I agree that these potions shouldn't be cheap, since they're quite strong, but the current cost is also just too grindy. If someone wanted to get a lot of them, they eventually would, but at the cost of a lot of wasted time. Therefore, I wouldn't make the potions as cheap as the vanilla ones, but I would make them require a more "normal" amount of reagents.

Here are the exact changes:

- Long potions: 32/16 redstone blocks —> 8/4
- Strong potions: 64/32 glowstone blocks —> 12/6
- Nausea potion: 32/16 mushrooms —> 8/4
- Blindness potion: 64/32 ink sacs —> 12/6
- Health boost potion: 64/32 —> 16/8

### Inventory

Vampirism potions are now also sorted after their vanilla variants. So a very long regeneration potion now comes after the other regeneration potions.

Screenshot 2026-09-01 at 12 12 58

### Other

There were also some issues with the new extended potions, mainly typos:

- The very strong instant health potion has an amplifier of 2, and so did its extended variant, which made them identical. Changed to 3.
- If I understand correctly, very strong variants of strong vanilla potions normally keep their duration but raise the amplifier by one. I'm not sure whether this was intentional, but very strong swiftness had a lower value than the original potion. Changed that as well.
- `VERY_LONG_STRENGTH_EXTENDED` was used as the extended variant of `VERY_STRONG_POISON`.

## Potion of Oblivion

- Renamed "Oblivion Potion" to "Potion of Oblivion" to match the other potions (yes, it isn't an ordinary potion, but it is still a potion — this is just a parity thing).
- Changed its texture to include an oblivion swirl.

Screenshot 2026-09-01 at 12 12 52

## ~~Dracula's~~ Sovereign Blood

I renamed Dracula's blood to sovereign blood. It sounds cooler, at least to me. I also made a new texture for it in the process and moved the item so that it comes after pure blood.

Screenshot 2026-09-01 at 12 15 42

I decided to go with a raspberry pink color for sovereign blood to make it easier to distinguish. The blood itself isn't the only part that is pink now — its name is too.

Screenshot 2026-09-01 at 12 18 03

I've added a separate rarity for such items — `vampirism:sovereign`. Currently it's used only for the blood itself, although the purity 6 tooltips on the respective items now have the same color as well. Normal pure blood items are marked with the uncommon rarity so that their names are golden, matching their textures.

Speaking of such items, I added the refined iron block to the creative tab, as it was missing there.

Screenshot 2026-09-01 at 12 27 23

## Strong Oils

I've removed the glint from strong oils and gave them a separate texture instead. They now have a raspberry pink ribbon. In addition, I've remade the normal oil textures on top of that.

Screenshot 2026-09-01 at 12 30 25

As you may notice in this screenshot, I've also changed the color of the activated and Dracula oils to match the blood colors. The latter was renamed to sovereign oil to match its blood type.

Also raised their max stack size to 16. Unlike potions, they don't have to be renewed often, so I suppose this should make them more convenient to use.

### Effects

I noticed that the effects of the potions and their strong alternatives were identical. The only real difference was the number of charges per bottle, which is... boring, especially as the only reward for defeating Dracula. This is why I made the effect amplifier higher for the stronger potions, which is what the prefix "strong" suggests in the first place. I'm not sure about lowering the effect duration... but if it were lowered, the potions might end up not being worth using, so I suppose it's better to leave it as it is.

### Tooltips

I basically just adjusted the tooltips for both the oils and the items they are applied to. They should be cleaner and more direct now, and they also show the number of uses left.

Screenshot 2026-09-02 at 22 50 08

Screenshot 2026-09-02 at 22 52 17

> Note: I plan to open a separate PR to clean up these huge tooltips. By that I mean hiding the faction restrictions when they are already met. The player already knows that they can wield a hunter axe, so there's no point in showing them this information.

### JEI

The oil application and removal recipes took up quite a lot of space in the inventory, so I condensed them into two recipes per oil-applicable item. This is what it looks like now. The removal recipe also includes the paper required to get the oil off the item.

Screenshot 2026-09-02 at 22 46 56

### Marshal Skills

The master alchemist and ultimate brewer skills already existed, but they weren't actually used anywhere. I've added them as requirements for making oils and brewing potions, so this is now fixed.


Ede1vveiss:
I like the changes, although as I've been playing hunter more this year. I would say making the potions _at least_ 4 times cheaper seems like a bit too much.
Perhaps if it was a making it 2 time cheaper than what it currently is then I think it should be fine. 

Also, I had an idea for the oils.
An extra durability indicator for how many uses the oils have left, that way it can be seen in the hotbar and you don't have to count or guess when the oil has run out. 
Here's an example of what I thought could be cool
Screenshot 2026-09-03 151516


Or perhaps have something similar to vampires with vampire swords having a separate durability bar.
image


_edit: misscounted how many times cheaper some of the potions were made_

TheGridExpert:
Screenshot 2026-09-03 at 19 18 56

The bar is already a thing, though I feel like it's too huge. I'd rather add an overlay over the item itself as well. 

As for potion costs... I know now they're too low on the contrast with the old ones, but I feel like lowering them by half is not enough. I'd do something inbetween:

Long potions:
- Initially: 32/16
- First edit: 8/4
- New edit: 12/6

Strong potions:
- Initially: 64/32
- First edit: 12/6
- New edit: 16/8

Ede1vveiss:
Are you aware of how easy it is to get redstone and glowstone nowadays?
Clerics also give redstone and glowstone and the new cave system are very favorable. If you want those ressources to be easier to get then just add those to the things minions can get.
I'd say even lowering it to half of what it is should already be enough.

I do agree that the current UI element is a bit too big. If it's going to stay, it would be great to be able to edit where it is and how big it is

TheGridExpert:
Okay, fine.

TheGridExpert:
## Potions

(Wanted to write an addition elaboration, but it turned into The Lord of the Rings, I hope it's fine)

I know that I have already agreed, but it's still stirring me. You're right about the cost of resources, they are indeed not *as* hard to get. It's not actually that expensive when you know what you're farming it for, but there is actually another reason why lowering them matters that I didn't mention (I somehow forgot about this after adding other stuff to the PR).

My initial concern was that the brewing mechanic is unpopular among people and this was supposed to make it a bit more interestingly-looking. When a player opens JEI the first thing they see now is "32 glowstone blocks to brew 5 potions". That's 288 dust against vanilla's 1. It is indeed not as much when we look at all the ways of obtaining such things in survival, but it looks absurd at first glance in comparison to normal potions that require one handful of glowstone dust. The player might just decide that this mechanic is not for them and never look at it again. A cost can be technically easy to pay and still never get paid, because the player never reaches the paying part with the mechanic becoming obsolete for them.

An additional factor that only adds to all of this is that, despite there being a skill that lets you cut ingredients by half, the first recipe that players see is the expensive one. If we take the second smaller number of resources as the base for calculations, we make the swift brewing skill unworthy of using (as it technically saves less time than the cost-efficient one does despite the name). So both skills should add convenience, but the first one is better not to be something vital or else it would eclipse the swift brewing completely.

I don't have any actual data to say that people outside of Vampirism.co barely use potions, but from what I've seen it's normally the minority, despite this mechanic being what carries hunters in fights. So, if we balance factions while considering potions, we should also make sure that they're actually used by players like we expect them to. If we do want data we can make a poll that would sound something like "how often do you brew hunter potions?" just to understand the numbers, but what I would like to lean towards to is convenience, not just cheapness. I know that this sounds like we're on some market, but players are generally easy, so this help it.

While pondering about this, I've got another idea. I could make it possible to brew vanilla brewing recipes on the vapor still. It removes the need in transfering bottles back and forth and also gives you two extra bottles of potions (provided you have the skill for that). This is not just for the value, though it's quite profitable on its own, but also remove that sixth potion when brewing via two batches on the vanilla brewing table. Could be convenient.

Also, fortune doesn't work on glowstone properly, unlike redstone, so I'd rather make recipes require identical amount of redstone and glowstone. Preferrably the lowest value between the current ones — 16/8 (I'd lean into a lower one, but at least not 32).

## Oils

Tried making a different display of the remaining oil. What do you think?

Screenshot 2026-09-04 at 00 11 48

Basically just a second bar above the durability one, but a blue one.

Ede1vveiss:
The Oils Durability looks good. 

For the potions, you can actually brew vanilla potions in the brewing / vapor still, it just isn't shown in JEI for some reason.

So for the costs of potions. I'd like to propose an alternative to just making potions cheaper. Letting hunters gather potion materials via minion gathering. Which I've explained in another post, feature #1638 .
I'd like to think that this would be a more interesting way to solving the convenience issue you talked about.

Cheaterpaul:
Look good. Nice additions.

## Oil Overlays

Not sure about the item overlay, but I would not necessarily add another type of overlay when we have the vampire sword one as well. Maybe can just shrink the existing one.

## Minions Tasks

We should to the suggestions regardless of this PR

## Potions

I have no opinion here. Though I would not make it required for player to use minions for the potions (they can help or make it easier, but it should still be possible without).

Maybe the better solutions is to make them cheaper and less effective at the same time. At the moment the timers are around 10x as long as the vanilla potions.
If it only about duration, cheaper recipes should be fine. If its about strength, that is something different 

---

Btw. JEI shows the same recipe but with 16 and 8 red stone blocks. Not sure if what happened there.
image

TheGridExpert:
Okay, so... quite a few changes so far.

## Charge Bars

Screenshot 2026-09-10 at 12 30 01

I've finished adding those bars. Now it's possible to choose between them and the previously existing ones in the client config, and their colors are also configurable.

colourwheel01

Oils lay a blue bar; it used to be purple, but I decided to use a more hunter-associated color. Initially, I supposed green would work better as it fit the theme of potions; however, it's already taken by durability, and thence it would have looked bad. As for vampire swords, I came up with red for blood charge and dark green for training. The training bar disappears as soon as it is filled or before it's first trained.

I decided not to add separate color configs for all stages as it would clutter the screen; that's why they're calculated via the color wheel. Basically, a util shifts the colors counter-clockwise as you can see on the screenshot above and on the right.

Also, those bars are registered through FactionApi so that, if other addons wanted to add a few, they wouldn't collide.

## Potions

I have completely recalculated very long and very strong potions; now the numbers are less arbitrary than they used to be. Their new duration is smaller, so this might be considered a nerf, but I feel this way, it might look more consequential.

Here's a table of durations and amplifiers for every potion, by effect and kind. The extended/sovereign variants, if they exist, have their duration written in parentheses. The ones added by the mod are in bold (except for the extended variants).

| Effect | Normal | Long | Strong | Very long | Very strong | Long + strong |
|---|---|---|---|---|---|---|
| Slow Falling | 1:30 | 4:00 | — | **11:00** (20:00) | — | — |
| Luck | 5:00 | **14:00** | — | **40:00** (60:00) | — | — |
| Weakness | 1:30 | 4:00 | — | **11:00** (20:00) | — | — |
| Strength | 3:00 | 8:00 | 1:30 II | **25:00** (40:00) | — | **11:00 II** (20:00 II) |
| Regeneration | 0:45 | 1:30 | 0:22 II | **3:00** (5:00) | **0:22 III** (0:36 III) | **1:30 III** (2:30 III) |
| Poison | 0:45 | 1:30 | 0:21 II | **3:00** (5:00) | **0:21 III** (0:34 III) | **1:30 III** (2:30 III) |
| Instant Health | instant | — | instant II | — | **instant III** (instant IV) | — |
| Water Breathing | 3:00 | 8:00 | — | **25:00** (40:00) | — | — |
| Slowness | 1:30 | 4:00 | 0:20 IV | **11:00** (20:00) | **0:20 VI** (0:32 VI) | **2:30 VI** (4:00 VI) |
| Swiftness | 3:00 | 8:00 | 1:30 II | **25:00** (40:00) | **1:30 III** (2:30 III) | **11:00 III** (20:00 III) |
| Fire Resistance | 3:00 | 8:00 | — | **25:00** (40:00) | — | — |
| Leaping | 3:00 | 8:00 | 1:30 II | **25:00** (40:00) | **1:30 III** (2:30 III) | **11:00 III** (20:00 III) |
| Invisibility | 3:00 | 8:00 | — | **25:00** (40:00) | — | — |
| Night Vision | 3:00 | 8:00 | — | **25:00** (40:00) | — | — |
| Wind Charged | 3:00 | **8:00** | — | **25:00** (40:00) | — | — |
| Weaving | 3:00 | **8:00** | — | **25:00** (40:00) | — | — |
| Oozing | 3:00 | **8:00** | — | **25:00** (40:00) | — | — |
| Infested | 3:00 | **8:00** | — | **25:00** (40:00) | — | — |
| Nausea | **1:30** | **4:00** | — | **11:00** (20:00) | — | — |
| Blindness | **1:30** | **4:00** | — | **11:00** (20:00) | — | — |
| Health Boost | **1:30** | **4:00** | **0:20 II** | **11:00** (20:00) | **0:20 III** (0:32 III) | **2:30 III** (4:00 III) |

The duration is calculated in the following way:

The game divides the duration of the long potion by the base duration, taking this as a factor. This factor is later raised to the power of 2 and multiplied by the base duration, and then the result is rounded. It might be clearer with an example:

Base: 3 minutes
Long: 8 minutes
8 / 3 = 2.6(6)
2.6(6) ^ 2 = 7.1(1)
3 * 7.1(1) = 21.3(3) minutes

The rounding is exact for small numbers, but larger for bigger ones; that's why it turns into 25 minutes.

The sovereign variants have the normal variant durations multiplied by 1.6. This multiplier is applied first, then the rounding.

## JEI

Screenshot 2026-09-12 at 16 10 51

- Expensive and efficient recipes of the same potions are now united into one box in order not to clutter the screen. The ingredients just scroll like other JEI recipes.
- The vapor still is now marked as a catalyst for brewing normal potions, and the vanilla brewing tab also gets opened when you click "Show Recipes" on the screen of the vapor still. This should make it clearer to players that you can brew vanilla potions there too.
- The "Req." plaque now demonstrates whether you have the skills required for this potion by its color.

projects/faction/src/main/java/de/teamlapen/faction/client/config/ExtendedConfigSectionScreen.java
```diff
@@ -0,0 +1,86 @@
...
+/**
+ * An extended variant of the config screen that allows translating and giving tooltips to enum values that extend {@link TranslatableEnum}
+ * <p>
+ * The translation keys have the following pattern:<br>
+ * "{mod_id}.configuration.{config_name}.{enum_name_lowercase}"
+ * <p>
+ * And tooltips have the same one but with a ".tooltip" suffix:<br>
+ * "{mod_id}.configuration.{config_name}.{enum_name_lowercase}.tooltip"
+ */
+public class ExtendedConfigSectionScreen extends ConfigurationScreen.ConfigurationSectionScreen {
```
Cheaterpaul: Why is this needed, the `TranslatableEnum#getTranslatedName` can be overridden in the enum to change the display text. And the description can be all displayed at the same time

projects/faction/src/main/java/de/teamlapen/faction/client/core/ItemBars.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.TreeMap;
+import java.util.function.Consumer;
+
+/**
+ * Draws bars over item slots, stacked upwards from the vanilla durability bar. All are sorted by their {@link Identifier},
+ * and the bars of a single provider keep the order they are handed to the consumer, bottom first.
+ */
+public class ItemBars {
```
Cheaterpaul, line 23: Can you move the `RegisterItemBarEvent`, `ItembarProvider` and `ItemBar` into the api.

Can you make `ItemBars` a field in `ClientServices` so most of the stuff can be non static

Cheaterpaul (changes requested):

Cheaterpaul (approved):

